PR #5, opened 2021-02-18
Specify error message for unmatched sells and avoid exception propagation from BilanceQueue

shredEngineer:
Even more specific error message on unmatched sells; replaced corresponding exception propagation.

### DIFF
--- a/src/bilance_queue.py
+++ b/src/bilance_queue.py
@@ -17,7 +17,7 @@
 import collections
 import dataclasses
 import logging
-from typing import Deque, Union
+from typing import Deque, Union, Optional
 import queue
 
 from transaction import *
@@ -58,19 +58,23 @@ class BilanceQueue:
         for fee in buffer_fee:
             self.remove_fee(fee)
 
-    def get(self) -> BilancedOperation:
+    def get(self) -> Optional[BilancedOperation]:
         """Get an item from the queue.
 
         Returns:
             BilancedOperation
+            ...or None, if the queue ran out of items to sell.
         """
         return self._get()
 
     def _put(self, bop: BilancedOperation) -> None:
         self.queue.append(bop)
 
-    def _get(self) -> BilancedOperation:
-        return self.queue.popleft()
+    def _get(self) -> Optional[BilancedOperation]:
+        try:
+            return self.queue.popleft()
+        except IndexError:
+            return None
 
     def remove_fee(self, fee: float) -> None:
         """Remove fee from the last added transaction.
@@ -99,7 +103,7 @@ class BilanceQueue:
             else:
                 fee -= not_sold
 
-    def sell(self, change: float) -> list[SoldCoin]:
+    def sell(self, change: float) -> Optional[list[SoldCoin]]:
         """Sell/remove coins from the queue, returning the sold coins.
 
         Depending on the QueueType, the coins will be removed FIFO or LIFO.
@@ -111,11 +115,15 @@ class BilanceQueue:
         Returns:
             list[SoldCoin]: List of specific coins which were (depending on
                             the tax regulation) sold in the transaction.
+            ...or None, if the queue ran out of items to sell.
         """
         assert change > 0
         sold_coins: list[SoldCoin] = []
         while True:
-            bop: BilancedOperation = self.get()
+            bop: Optional[BilancedOperation] = self.get()
+
+            if bop is None:
+                return None
 
             not_sold = bop.op.change - bop.sold
             assert not_sold > 0
--- a/src/book.py
+++ b/src/book.py
@@ -50,7 +50,8 @@ class Book:
             # Skip header.
             next(reader)
 
-            for row, (_utc_time, account, operation, coin, _change, remark) in enumerate(reader, 2):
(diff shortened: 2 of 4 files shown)


src/bilance_queue.py
```diff
@@ -58,19 +58,23 @@ def put(self, item: Union[Operation, BilancedOperation]) -> None:
         for fee in buffer_fee:
             self.remove_fee(fee)
 
-    def get(self) -> BilancedOperation:
+    def get(self) -> Union[BilancedOperation, None]:
```
provinzio, line 61: We should use Optional[object] instead of Union[object, None].

src/book.py
```diff
@@ -83,7 +83,7 @@ def _read_binance(self, file_path: Path) -> None:
                         "Could not recognize operation `%s` in binance file `%s:%i`.", operation, file_path, row)
                     continue
 
-                o = Op(utc_time, platform, change, coin)
+                o = Op(utc_time, platform, change, coin, reader.line_num)
```
provinzio, line 86: Oh nice, I didn't know about `line_num`. We should replace the already used `row` variable with this one. Then we no longer need enumerate.

We cold even start with `row = reader.line_num` to keep the variable short and handy and remove enumerate.

src/taxman.py
```diff
@@ -77,13 +77,18 @@ def _evaluate_taxation_GERMANY(self, coin: str, operations: list[Operation]) ->
...
-                    if coin == config.FIAT:
+                sold_coins = bilance.sell(op.change)
+                if sold_coins is None:          # Queue ran out of items to sell...
+                    if coin == config.FIAT:     # ...this is OK for fiat currencies (not taxable)
                         continue
-                    raise RuntimeError(
-                        f"Not enough {coin} in queue to sell (transaction from {op.utc_time} on {op.platform}).\nHave you forgotten to add all account statements?\nThis error could occure after deposits from unknown sources.")
+                    else:                       # ...but not for crypto coins (taxable)
+                        log.error(
+                            f"Line {op.line}: "
```
provinzio, line 86: We could save the file with the line number altogether to keep it explicit.

src/taxman.py
```diff
@@ -77,13 +77,18 @@ def _evaluate_taxation_GERMANY(self, coin: str, operations: list[Operation]) ->
             elif isinstance(op, Buy):
                 bilance.put(op)
             elif isinstance(op, Sell):
-                try:
-                    sold_coins = bilance.sell(op.change)
-                except IndexError:
-                    if coin == config.FIAT:
+                sold_coins = bilance.sell(op.change)
+                if sold_coins is None:          # Queue ran out of items to sell...
```
provinzio, line 81: I'd like to stick to autopep8 for formatting (using 2 spaces before the inline comment). Otherwhise my autoformatter will change it after my next commit :D

Depending on your IDE you could use to one. I would recommend using an autoformatter. Otherwhise ignore this comment.

We may even want to avoid inline comments and move them in their resepctive if/else-clause.

provinzio (commented):
Hey @shredEngineer,

thank you for your participation and time. Saving the line number from where the Operations come, is a good idea.
I added my two cents in your code.

Looking forward to your Kraken integration.

src/bilance_queue.py
shredEngineer: Alright. I have to admit I don't use type annotations that often. Learning never stops! :)

src/taxman.py
shredEngineer: Thanks for the info, I'll stick to that!

src/book.py
```diff
@@ -30,6 +30,7 @@ class Book:
 
     def __init__(self) -> None:
         self.operations: list[tr.Operation] = []
+        self.current_file_path = None
```
provinzio, line 33: I think, this does not work. The Book keeps record of all operations from all files. I guess we have to save the file path to the Operations class (like the line number).

src/book.py
```diff
@@ -50,7 +51,8 @@ def _read_binance(self, file_path: Path) -> None:
             # Skip header.
             next(reader)
 
-            for row, (_utc_time, account, operation, coin, _change, remark) in enumerate(reader, 2):
+            for (_utc_time, account, operation, coin, _change, remark) in reader:
```
provinzio, line 54: remove unnecessary brackets

provinzio (changes requested):

src/book.py
provinzio: I am currently thinking of something like.. but this could be overkill. I am unsure if this is smart.

But going this explicit, we could implement __str__ on FileLocation and print the file location directly with log.warning()

```
@dataclass
class FileLocation:
  file: str
  line: int

  def __str__(self) -> str:
    return f"{self.file}:{self.line}"

class Operation:
  ...
  fl: FileLocation
```

shredEngineer: Oh, I didn't think of that. In that case, we should really make the full file path a member of every `Operation`. I guess there won't be millions of operations, so we shouldn't have to worry about this waste of memory.

(In `Book`, we could also keep a dictionary of file paths and IDs, assigning only the corresponding ID to `Operation` -- but I think that's overkill for now.)

provinzio (approved):

PR #7, opened 2021-02-19
Implement Exchange coinbase

### LINKED ISSUE #3: Implement Exchange coinbase
Account Statements CSV looks easy to be parsed.

API can be accessed without API KEY: https://developers.coinbase.com/api/v2#prices

provinzio:
Closes #3

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Pull Requests und Anfragen über Issues sind gerne gesehen (siehe `Key notes for
 
 ### Currently supported exchanges
 - Binance
+- coinbase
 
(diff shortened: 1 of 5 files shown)


shredEngineer:
Code looks good!

But I have to wonder if it's actually worth the struggle to redundantly store the prices in the database, whereas these are contained in the CSV file anyway. I am under the impression that the database is really only needed to reduce networking (API) requests, which is not required in this case.

provinzio:
@shredEngineer Feel free to use the new `Book.append_operations` function on your branch.

PR #9, opened 2021-02-23
ADD Exchange Kraken

### LINKED ISSUE #8: Implement exchange Kraken
For Kraken, there are two [different](https://support.kraken.com/hc/en-us/articles/115000302707-Differences-between-ledger-and-trades-history) CSV export options for account statements: _Trades_ and _Ledgers_ history.

We need to use the _Ledgers_ history, as its format is vastly more useful for our purposes.

The transaction fees are listed in-line (and not in separate follow-up lines), so I think for each line the read-in function will have to generate _two_ `Operation`s (in the case of a non-zero fee).

Historical prices (OHLC data) and trade data can easily be fetched using the public API: https://support.kraken.com/hc/en-us/articles/218198197-How-to-retrieve-historical-time-and-sales-trading-history-using-the-REST-API-Trades-endpoint-

I'm currently working on this feature.

shredEngineer:
Closes #8

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -24,6 +24,7 @@ import config
 import misc
 from price_data import PriceData
 import transaction as tr
+from core import kraken_asset_map
 
 log = logging.getLogger(__name__)
 
@@ -179,6 +180,79 @@ class Book:
                     self.append_operation("Fee", utc_time, platform,
                                           eur_fee, "EUR", row, file_path)
 
+    def _read_kraken_trades(self, file_path: Path) -> None:
+        log.error(f"{file_path.name}: "
+                  f"Looks like this is a Kraken 'Trades' history, but we need the 'Ledgers' history. "
+                  f"(See: https://github.com/provinzio/CoinTaxman/wiki/Exchange:-Kraken)")
+
+    def _read_kraken_ledgers(self, file_path: Path) -> None:
+        platform = "kraken"
+        operation_mapping = {
+            "spend": "Sell",  # Sell ordered via 'Buy Crypto' button
+            "receive": "Buy",  # Buy ordered via 'Buy Crypto' button
+            "transfer": "Airdrop",
+            "reward": "StakingInterest",
+            "deposit": "Deposit",
+            "withdrawal": "Withdraw",
+        }
+
+        # Need to track state of "duplicate entries" for deposits / withdrawals;
+        # the second deposit and the first withdrawal entry need to be skipped.
+        #   dup_state["deposit"] == 0: Deposit is broadcast to blockchain            <-- Taxable event (is in public trade history)
+        #   dup_state["deposit"] == 1: Deposit is credited to Kraken account         <-- Skipped
+        #   dup_state["withdrawal"] == 0: Withdrawal is requested in Kraken account  <-- Skipped
+        #   dup_state["withdrawal"] == 1: Withdrawal is broadcast to blockchain      <-- Taxable event (is in public trade history)
+        dup_state, dup_skip = {"deposit": 0, "withdrawal": 0}, {"deposit": 1, "withdrawal": 0}
+        # See: https://support.kraken.com/hc/en-us/articles/360001169443-Why-there-are-duplicate-entries-for-deposits-withdrawals
+
+        with open(file_path, encoding="utf8") as f:
+            reader = csv.reader(f)
+
+            # Skip header.
+            next(reader)
+
+            for txid, refid, _utc_time, _type, subtype, aclass, _asset, _amount, _fee, balance in reader:
+                row = reader.line_num
+
+                # Skip "duplicate entries" for deposits / withdrawals
+                if _type in dup_state.keys():
+                    skip = dup_state[_type] == dup_skip[_type]
+                    dup_state[_type] = (dup_state[_type] + 1) % 2
+                    if skip:
+                        continue
+
+                # Parse data.
+                utc_time = datetime.datetime.strptime(_utc_time, "%Y-%m-%d %H:%M:%S").replace(tzinfo=datetime.timezone.utc)
+                change = float(_amount)
+                coin = kraken_asset_map.get(_asset, _asset)
+                fee = float(_fee)
+                operation = operation_mapping.get(_type)
+                if operation is None:
+                    if _type == "trade":
+                        operation = "Sell" if change < 0 else "Buy"
+                    elif _type in ["margin trade", "rollover", "settled"]:
+                        log.error(f"{file_path}: {row}: Margin trading is currently not supported. Please create an Issue or PR.")
+                        raise RuntimeError
+                    else:
+                        log.error(f"{file_path}: {row}: Other order type '{_type}' is currently not supported. Please create an Issue or PR.")
+                        raise RuntimeError
+                change = abs(change)
+
+                # Validate data.
+                assert operation
+                assert coin
+                assert change
+
+                self.append_operation(operation, utc_time, platform,
+                                      change, coin, row, file_path)
+
+                if fee != 0:
+                    self.append_operation("Fee", utc_time, platform,
+                                          fee, coin, row, file_path)
+
+        assert dup_state["deposit"] == 0, "Orphaned deposit. (Must always come in pairs). Is your file corrupted?"
+        assert dup_state["withdrawal"] == 0, "Orphaned withdrawal. (Must always come in pairs). Is your file corrupted?"
+
     def detect_exchange(self, file_path: Path) -> Optional[str]:
         if file_path.suffix == ".csv":
             with open(file_path, encoding="utf8") as f:
@@ -189,6 +263,8 @@ class Book:
                 "binance": ['UTC_Time', 'Account',
                             'Operation', 'Coin', 'Change', 'Remark'],
                 "coinbase": ['You can use this transaction report to inform your likely tax obligations. For US customers, Sells, Converts, and Rewards Income, and Coinbase Earn transactions are taxable events. For final tax obligations, please consult your tax advisor.'],
+                "kraken_ledgers": ["txid", "refid", "time", "type", "subtype", "aclass", "asset", "amount", "fee", "balance"],
+                "kraken_trades": ["txid", "ordertxid", "pair", "time", "type", "ordertype", "price", "", "cost", "fee", "vol", "margin", "misc", "ledgers"],
             }
             for exchange, expected in expected_headers.items():
                 if header == expected:
--- a/src/core.py
+++ b/src/core.py
@@ -191,3 +191,38 @@ class Fiat(Enum):
     ZAR = "South Africa Rand"
     ZMW = "Zambia Kwacha"
     ZWD = "Zimbabwe Dollar"
+
+# Kraken's "dirty" asset codes don't seem to follow any system. Only "clean" names are used in CoinTaxman internally.
+#     kraken_asset_map:  Converts Kraken "dirty" asset codes to "clean" asset names           (e.g. ZEUR   -> EUR)
+#     kraken_pair_map:   Converts clean fiat / clean crypto pairs to "dirty" API asset pairs  (e.g. ETHEUR -> XETHZEUR)
+# Analyzed using asset pairs API data: https://api.kraken.com/0/public/AssetPairs (retrieved at 2021-02-21)
+# … because this is deprecated: https://support.kraken.com/hc/en-us/articles/360001185506-Explanation-of-Asset-Codes
+# See also: https://support.kraken.com/hc/en-us/articles/202944246-All-available-currencies-and-trading-pairs-on-Kraken
+kraken_asset_map = {
+    # Fiat:
+    "ZUSD": "USD",
+    "ZEUR": "EUR",
+    "ZCAD": "CAD",
+    "ZJPY": "JPY",
+    "ZGBP": "GBP",
+    "CHF" : "CHF",
+    "ZAUD": "AUD",
+    # Crypto:
+    "XXBT": "XBT",
+    "XETH": "ETH",
+    "XLTC": "LTC",
+    "XXDG": "XDG",
+    "XXRP": "XRP"
+}
+kraken_pair_map = {  # Only these asset pairs violate the rule "clean name" + "clean name" = "asset pair"
+    'USDTUSD': 'USDTZUSD', 'XETCETH': 'XETCXETH', 'XETCXBT': 'XETCXXBT', 'XETCEUR': 'XETCZEUR', 'XETCUSD': 'XETCZUSD',
+    'ETHXBT': 'XETHXXBT', 'ETHCAD': 'XETHZCAD', 'ETHEUR': 'XETHZEUR', 'ETHGBP': 'XETHZGBP', 'ETHJPY': 'XETHZJPY',
+    'ETHUSD': 'XETHZUSD', 'LTCXBT': 'XLTCXXBT', 'LTCEUR': 'XLTCZEUR', 'LTCJPY': 'XLTCZJPY', 'LTCUSD': 'XLTCZUSD',
+    'XMLNETH': 'XMLNXETH', 'XMLNXBT': 'XMLNXXBT', 'XMLNEUR': 'XMLNZEUR', 'XMLNUSD': 'XMLNZUSD', 'XREPETH': 'XREPXETH',
+    'XREPXBT': 'XREPXXBT', 'XREPEUR': 'XREPZEUR', 'XREPUSD': 'XREPZUSD', 'XBTCAD': 'XXBTZCAD', 'XBTEUR': 'XXBTZEUR',
+    'XBTGBP': 'XXBTZGBP', 'XBTJPY': 'XXBTZJPY', 'XBTUSD': 'XXBTZUSD', 'XDGXBT': 'XXDGXXBT', 'XXLMXBT': 'XXLMXXBT',
+    'XXLMAUD': 'XXLMZAUD', 'XXLMEUR': 'XXLMZEUR', 'XXLMGBP': 'XXLMZGBP', 'XXLMUSD': 'XXLMZUSD', 'XXMRXBT': 'XXMRXXBT',
+    'XXMREUR': 'XXMRZEUR', 'XXMRUSD': 'XXMRZUSD', 'XRPXBT': 'XXRPXXBT', 'XRPCAD': 'XXRPZCAD', 'XRPEUR': 'XXRPZEUR',
+    'XRPJPY': 'XXRPZJPY', 'XRPUSD': 'XXRPZUSD', 'XZECXBT': 'XZECXXBT', 'XZECEUR': 'XZECZEUR', 'XZECUSD': 'XZECZUSD',
+    'EURUSD': 'ZEURZUSD', 'GBPUSD': 'ZGBPZUSD', 'USDCAD': 'ZUSDZCAD', 'USDJPY': 'ZUSDZJPY'
+}
--- a/src/price_data.py
+++ b/src/price_data.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import bisect
 import datetime
 import json
 import logging
@@ -24,6 +25,7 @@ from typing import Optional, Union
 import requests
 
 import config
+from core import kraken_pair_map
 import misc
 from transaction import *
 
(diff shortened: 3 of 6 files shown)


src/book.py
```diff
@@ -16,6 +16,7 @@
 
 import csv
 import datetime
+from dateutil.tz import tzutc
```
shredEngineer, line 19: I had to explicitly set the timezone to UTC on my timestamps. I guess you should also check if your timestamps match what you expect.

src/core.py
```diff
@@ -191,3 +191,39 @@ class Fiat(Enum):
     ZAR = "South Africa Rand"
     ZMW = "Zambia Kwacha"
     ZWD = "Zimbabwe Dollar"
+
```
shredEngineer, line 194: I created these mappings using a simple python script which fetches the currently supported AssetPairs JSON and checks which pairs need "special attention" as per the "kraken_pair_map".

[Kraken-AssetPairs-Parse.py.txt](https://github.com/provinzio/CoinTaxman/files/6028235/Kraken-AssetPairs-Parse.py.txt)

src/price_data.py
```diff
@@ -26,8 +26,10 @@
 import config
 import misc
 from transaction import *
+from core import kraken_pair_map
 
 log = logging.getLogger(__name__)
+logging.getLogger("urllib3").propagate = False
```
shredEngineer, line 32: You can decide if you want to keep those urllib debug lines in the log. I found them quite distracting, so this is the code to disable these.

src/price_data.py
```diff
@@ -104,6 +106,60 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+            # Find closest timestamp match
+            data = data["result"][pair]
+            closest_match_index = None
+            for i, d in enumerate(data):
+                trade_time = int(float(d[2]) * 1000)
+                time_diff = target_time - trade_time
+                if time_diff >= 0:
+                    closest_match_index = i
+                else:
+                    break
```
provinzio, line 153: I believe that the timestamps are sorted? Lets use [bisect](https://docs.python.org/3/library/bisect.html) here.
`closest_match_index = bisect.bisect_left(data, utc_time) - 1` should do the job.

src/price_data.py
```diff
@@ -104,6 +106,60 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+
+            if closest_match_index is not None:
```
provinzio, line 155: Using bisect, this has to be `if closest_match_index > 0:`

src/price_data.py
shredEngineer: Absolutely. :+1: I seem to have overlooked that file altogether…

shredEngineer:
> Hey @shredEngineer,
> 
> you did a nice job implementing the Kraken API to CoinTaxman. I am unfortunatly unable to test it myself, because I have no test files.
> 
> I added my thoughts to your code. I'd love if you could revisit your code with them.

Thank you! :) Will do.

In the meantime, here is an example account statement (matching ledgers and trades) I just prepared: 
[kraken-example-ledgers.csv.txt](https://github.com/provinzio/CoinTaxman/files/6068419/kraken-example-ledgers.csv.txt)
[kraken-example-trades.csv.txt](https://github.com/provinzio/CoinTaxman/files/6068420/kraken-example-trades.csv.txt)

src/book.py
```diff
@@ -176,6 +178,79 @@ def _read_coinbase(self, file_path: Path) -> None:
...
 
+    def _read_kraken_trades(self, file_path: Path) -> None:
+        log.error(f"{file_path.name}: "
+                  f"Looks like this is a Kraken 'Trades' history, but we need the 'Ledgers' history. "
+                  f"(See: https://github.com/provinzio/CoinTaxman/wiki/Exchange:-Kraken)")
+
+    def _read_kraken_ledgers(self, file_path: Path) -> None:
+        platform = "kraken"
+        operation_mapping = {
+            "trade": None,  # Buy or Sell order
```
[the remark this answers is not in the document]
shredEngineer: I included it just for explicitness, so it's obvious to the reader which operations are supported at all. But you're right, implementation-wise it doesn't make any sense. I will remove it.

src/book.py
```diff
@@ -176,6 +178,79 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+
+                # Skip "duplicate entries" for deposits / withdrawals
+                if _type in dup_state.keys():
+                    dup_state[_type] = (dup_state[_type] + 1) % 2
+                    if dup_state[_type] == (dup_skip[_type] + 1) % 2:
+                        continue
+
+                # Parse data.
+                utc_time = datetime.datetime.strptime(_utc_time, "%Y-%m-%d %H:%M:%S")
+                utc_time = utc_time.replace(tzinfo=tzutc())
```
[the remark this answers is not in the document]
shredEngineer: Got it!

src/price_data.py
```diff
@@ -104,6 +106,60 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+        target_time = int(utc_time.timestamp() * 1000)
+        root_url = "https://api.kraken.com/0/public/Trades"
+        pair = base_asset + quote_asset
+        pair = kraken_pair_map.get(pair, pair)
+
+        minutes_offset = 0
+        while minutes_offset < 120:
+            minutes_offset += 10
+
+            start_time, end_time = misc.get_offset_timestamps(utc_time, datetime.timedelta(minutes=minutes_offset))
```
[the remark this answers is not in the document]
shredEngineer: Good idea; I've corrected this.

src/price_data.py
```diff
@@ -104,6 +106,60 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+            since = start_time * 1000000
```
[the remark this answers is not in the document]
shredEngineer: > Why do we have to multiply?

The Kraken API is _really_ cumbersome. Timestamps are excepted in nanoseconds, thus the multiplication.

> And how is it possible to get from a timestamp to a trade id?

It is impossible. That why we need to scan for the closest timestamp before any particular order. I don't like it, and maybe I didn't understand it completely. But here's how I see it: We want to get the exchange rate of any particular asset at some particular time, without juggling the two matching Kraken account statement files at once.
And you seem to do a similar thing in the Binance case -- but you're averaging over some interval instead of finding the last exchange rate _before_ that particular trade.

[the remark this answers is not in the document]
provinzio: I asked about the trade id because the documentation says, that it takes a id for `since` instead of a timestamp. Or am I reading the documentation wrong?

> pair = asset pair to get trade data for
> since = return trade data since given id (optional.  exclusive)

https://www.kraken.com/features/api

src/price_data.py
```diff
@@ -104,6 +106,60 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+            if data["error"]:
+                log.warning(f"Querying trades for {pair} at {utc_time} (offset={minutes_offset}m): "
+                            f"Could not retrieve trades: {data['error']}")
+                return 0
+
+            # Find closest timestamp match
+            data = data["result"][pair]
+            closest_match_index = None
+            for i, d in enumerate(data):
+                trade_time = int(float(d[2]) * 1000)
```
provinzio, line 148: We could add a comment here `Convert to ms`. To keep it easy to understand in which unit the timestamp is.

src/price_data.py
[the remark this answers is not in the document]
shredEngineer: Hmm, I just tried it with a "txid" and an "ordertxid", both with and without specifying the underlying asset pair. But the parameter value is ignored! I guess the "ID" is really just the timestamp in nanoseconds. 

EDIT: Oh ok: `last = id to be used as since when polling for new trade data`. As this "last" value is clearly a timestamp, this leaves no doubt about the type of ID.

src/price_data.py
shredEngineer: Check

src/book.py
```diff
@@ -176,6 +178,79 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+                change = float(_amount)
+                coin = kraken_fiat_map.get(_asset, kraken_crypto_map.get(_asset, _asset))
```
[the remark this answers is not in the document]
shredEngineer: While in principle we could, I'd personally find that inelegant, since `kraken_fiat_map` maps a single asset, while `kraken_crypto_map` maps an asset-_pair_.

src/price_data.py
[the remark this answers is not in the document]
provinzio: I found an example [here](https://support.kraken.com/hc/en-us/articles/218198197-How-to-retrieve-historical-time-and-sales-trading-history-using-the-REST-API-Trades-endpoint-), which states that it is the timestamp in ns.

src/price_data.py
shredEngineer: Oh boy! Another new function for me. I can certainly look into that, but it'll take some time. I don't really see how that line would work, especially in the context of the for-loop.

src/book.py
```diff
@@ -176,6 +178,79 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+            "withdrawal": "Withdraw",
+        }
+
+        # Need to track state of "duplicate entries" for deposits / withdrawals:
+        #   dup_state["deposit"] == 1: Deposit is broadcast to blockchain            <-- Taxable event (is in public trade history)
+        #   dup_state["deposit"] == 2: Deposit is credited to Kraken account         <-- Skipped
+        #   dup_state["withdrawal"] == 1: Withdrawal is requested in Kraken account  <-- Skipped
+        #   dup_state["withdrawal"] == 2: Withdrawal is broadcast to blockchain      <-- Taxable event (is in public trade history)
+        dup_state, dup_skip = {"deposit": 0, "withdrawal": 0}, {"deposit": 2, "withdrawal": 1}
+        # See: https://support.kraken.com/hc/en-us/articles/360001169443-Why-there-are-duplicate-entries-for-deposits-withdrawals
```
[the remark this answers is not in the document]
shredEngineer: > Should it be "deposit": 1? Your calculations below never result in 2.

I guess I was lazy coding this. I'm comparing _after_ incrementing, thus saving a line or two of code, but instead "wrapping around" the value to compare with. I'll now make this very explicit, without using modulo.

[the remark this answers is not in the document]
shredEngineer: EDIT²: On a second thought, I implemented it like this in order to be able to select which event is the taxable one. I'm absolutely no expert on this. If you can confirm that the "broadcast to blockchain" event is the right one, I can hard-code this. Otherwise, lets leave it like that, using counters?!

src/price_data.py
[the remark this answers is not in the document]
shredEngineer: Yep, that should settle it. Damn! Just inserting a trade ID would've been much nicer.

src/book.py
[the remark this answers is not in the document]
provinzio: I believe that the broadcast to blockchain is the relevant event. The other seems to me internal only.

src/book.py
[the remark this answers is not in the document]
provinzio: I do not understand you. `kraken_fiat_map` maps fiat, `kraken_crypto_map` maps crypto. We could merge them. Both map the internal Kraken asset to the real asset. I would not change `kraken_pair_map`.

src/price_data.py
provinzio: You do not need a forloop for that. bisect is nothing else than a binary search which should be faster on sorted lists than checking each value.

By the way. We have to confirm, that at least the last value (the latest trade) is after our trade. Otherwhise we would get the wrong price.

src/book.py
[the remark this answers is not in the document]
shredEngineer: Sorry, I confused `kraken_crypto_map` with `kraken_pair_map`. Yes, I'll merge them.

src/book.py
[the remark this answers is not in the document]
shredEngineer: Ok, I thought so, too, but wasn't 100% sure about it.

But since we need to skip the _first_ deposit event, and need to skip the _second_ withdrawal event, I think using counters is still the most simple solution?

src/price_data.py
shredEngineer: Thanks for the clarification, I'll look into that tomorrow. Although speed shouldn't really be a big concern here, I'm always happy to learn about other ways to do things!

Yes, I guess you're right that there needs to be some extra logic handling the case of an "extremely dense" trade history where -10 minutes is already too far back.

src/book.py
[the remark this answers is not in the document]
provinzio: Well, I still like the boolean idea, which might be easier to understand? The current solution looks too complicated for the use case.

If you want to stand with your current implementation, you could change `(dup_skip[_type] + 1) % 2` to just `dup_skip[_type]` and change the skipping values to 0 and 1. Which removes one layer of complexity.

shredEngineer:
I may have screwed up the push, because I forgot to rebase on the stuff you commited in the meantime.
My days-old "ADD Exchange Kraken" commit now appears as if I had first pushed it a few minutes ago.
Can you work with this anyway? (Sorry!)

provinzio:
mhh interesting. `git rebase provinzio/main` shouldn't change the date of your commit. But it's no problem. I beliebve, that the commit still has the same changes. I'll only check your changes in ce47ba8

shredEngineer:
> mhh interesting. `git rebase provinzio/main` shouldn't change the date of your commit. But it's no problem. I believe, that the commit still has the same changes. I'll only check your changes in [ce47ba8](https://github.com/provinzio/CoinTaxman/commit/ce47ba85c57e652e836e5210bb70481338805a46)

I may have done some other hasty git stuff without thinking, sorry again; I'll be more careful next time. Glad this isn't a huge problem. :) Yes, the old commit itself is unaltered. You're safe to only check the new commit.

src/book.py
```diff
@@ -179,6 +180,77 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+            next(reader)
+
+            for txid, refid, _utc_time, _type, subtype, aclass, _asset, _amount, _fee, balance in reader:
+                row = reader.line_num
+
+                # Skip "duplicate entries" for deposits / withdrawals
+                if _type in dup_state.keys():
+                    dup_state[_type] = (dup_state[_type] + 1) % 2
+                    if dup_state[_type] == dup_skip[_type]:
+                        continue
```
provinzio, line 220: I think that the counter logic is still too cumbersome to read.

We may lighten this up by changing the order of `dup_state[_type] = (dup_state[_type] + 1) % 2` and the if-case.
We have to change the skip values to 1, 0 for this to work.

This leaves us with 0 -> first entry, 1 -> second entry. Currently it is 1 -> first entry, 0 -> second entry, which I find weird.

Changing this, we can order the state comments above consecutivly (0, 1) instead of (1,0)

src/config.py
```diff
@@ -24,7 +24,7 @@
 
 # User specific constants.
 COUNTRY = core.Country.GERMANY
-TAX_YEAR = 2020
+TAX_YEAR = 2021
```
provinzio, line 27: leave it at 2020

src/book.py
```diff
@@ -179,6 +180,77 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+            "deposit": "Deposit",
+            "withdrawal": "Withdraw",
+        }
+
+        # Need to track state of "duplicate entries" for deposits / withdrawals:
+        #   dup_state["deposit"] == 1: Deposit is broadcast to blockchain            <-- Taxable event (is in public trade history)
+        #   dup_state["deposit"] == 0: Deposit is credited to Kraken account         <-- Skipped
+        #   dup_state["withdrawal"] == 1: Withdrawal is requested in Kraken account  <-- Skipped
+        #   dup_state["withdrawal"] == 0: Withdrawal is broadcast to blockchain      <-- Taxable event (is in public trade history)
+        dup_state, dup_skip = {"deposit": 0, "withdrawal": 0}, {"deposit": 0, "withdrawal": 1}
```
provinzio, line 204: I'd really prefer a small comment here, saying that the Kraken statement has two entries for each deposit and withdrawl. The second deposit and first withdrawl entries are Kraken internal and will be skipped.

It's way easier to understand what you want to accomplish if it's there in text form.
Rows 200-203 can stay, but adding the text might help a lot.

src/book.py
```diff
@@ -179,6 +180,77 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+
+                self.append_operation(operation, utc_time, platform,
+                                      change, coin, row, file_path)
+
+                if fee != 0:
+                    self.append_operation("Fee", utc_time, platform,
+                                          fee, coin, row, file_path)
+
+        assert dup_state["deposit"] == 0, "Orphaned deposit. (Must always come in pairs). Is your file corrupted?"
+        assert dup_state["withdrawal"] == 0, "Orphaned withdrawal. (Must always come in pairs). Is your file corrupted?"
```
provinzio, line 252: If the change the calculation/if-case order, we have to adjust the assert statment to `==1`

src/price_data.py
```diff
@@ -26,6 +26,8 @@
 import config
 import misc
 from transaction import *
+from core import kraken_pair_map
+import bisect
```
provinzio, line 30: I like to order my imports in three categories which are ordered each alphabetically.

1. standardlib
2. third-party
3. internal

This would look like
```
import bisect

import config
from core import kraken_pair_map
import misc
from transaction import *
```

This helps to get a quick overlook over the imports in the file.
Just my taste, you do not have to bend for this ;)

src/price_data.py
```diff
@@ -104,6 +106,69 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
         average_price = total_cost / total_quantity
         return average_price
 
+    @misc.delayed
+    def _get_price_kraken(self, base_asset: str, utc_time: datetime.datetime, quote_asset: str, minutes_step: int = 10) -> float:
+        """Retrieve price from Kraken official REST API.
+
+        We select the data point closest to the desired timestamp (utc_time), but not newer than this timestamp.
+        For this we fetch one chunk of the trade history, starting 10 minutes before this timestamp.
```
provinzio, line 114: ... starting `minutes_step` before ...

src/price_data.py
```diff
@@ -104,6 +106,69 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+        We then walk through the history until the closest timestamp match is found.
+        Otherwise, we start another 10 minutes earlier and try again, etc. …
+        (Exiting with a warning and zero price after hitting the arbitrarily chosen offset limit of 120 minutes.)
+        If the 10 minutes-offset is already too large, recursively retry by reducing the offset step, down to 1 minute.
+
+        Documentation: https://www.kraken.com/features/api
+        """
```
provinzio, line 121: We should add args/returns documentation here.
The docstrings should be according to the [google style guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods). Have a look at the example on the google style guide page.

src/price_data.py
```diff
@@ -104,6 +106,69 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+    @misc.delayed
+    def _get_price_kraken(self, base_asset: str, utc_time: datetime.datetime, quote_asset: str, minutes_step: int = 10) -> float:
+        """Retrieve price from Kraken official REST API.
+
+        We select the data point closest to the desired timestamp (utc_time), but not newer than this timestamp.
+        For this we fetch one chunk of the trade history, starting 10 minutes before this timestamp.
+        We then walk through the history until the closest timestamp match is found.
+        Otherwise, we start another 10 minutes earlier and try again, etc. …
+        (Exiting with a warning and zero price after hitting the arbitrarily chosen offset limit of 120 minutes.)
+        If the 10 minutes-offset is already too large, recursively retry by reducing the offset step, down to 1 minute.
```
provinzio, line 118: If the initial offset is ...

src/price_data.py
```diff
@@ -104,6 +106,69 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+        target_timestamp = misc.to_ms_timestamp(utc_time)
+        root_url = "https://api.kraken.com/0/public/Trades"
+        pair = base_asset + quote_asset
+        pair = kraken_pair_map.get(pair, pair)
+
+        minutes_offset = 0
+        while minutes_offset < 120:
+            minutes_offset += minutes_step
+
+            since = misc.to_ms_timestamp(utc_time - datetime.timedelta(minutes=minutes_offset)) * 1000000
```
provinzio, line 131: Lets add a function `to_ns_timestamp`

src/price_data.py
```diff
@@ -104,6 +106,69 @@ def _get_price_binance(self, base_asset: str, utc_time: datetime.datetime, quote
...
+            data_timestamps_ms = [int(float(d[2]) * 1000) for d in data]
+            closest_match_index = bisect.bisect_left(data_timestamps_ms, target_timestamp) - 1
+
+            # The desired timestamp is in the past; increase the offset
+            if closest_match_index == - 1:
+                continue
+
+            # The desired timestamp is in the future.
+            # Ensure at least one additional value at the end of the array to be sure our match is plausible
+            if closest_match_index >= len(data_timestamps_ms) - 2:
```
provinzio, line 155: Shouldn't this be `>`?

In the case, that to many trades were made in the 10 minutes interval. `bisect_left` returns `len(data_timestamps_ms)`. this makes `closest_match_index = len(data_timestamps_ms) - 1`.

I think we can savely change it to if `closest_match_index == len(data_timestamps_ms) - 1`. Using >= is possible here, but unnecessary I guess.

provinzio (changes requested):

src/price_data.py
shredEngineer: Got it!

src/price_data.py
shredEngineer: Alright. (You should also complete the docstrings for `_get_price_binance`. :) )

provinzio: Thanks :) I added the missing documentation there (https://github.com/provinzio/CoinTaxman/commit/eee3f3ba4c98def7fae0497ce6c84e8d8745116a).

src/price_data.py
shredEngineer: You're right, I got confused with the idices.

src/config.py
shredEngineer: Yep, wasn't intended, it was just meant to be set locally for testing. Will change it back.

src/book.py
shredEngineer: Changed it accordingly.

src/book.py
shredEngineer: I don't think so! If deposits come in pairs, the counter will always be zero. Same for the withdrawals.

shredEngineer:
God dammit, I somehow messed up again. Please let's just get this over with. :rofl:

src/book.py
provinzio: You are right.

provinzio:
Have you cherry picked my change? :D Looks like it. I am really wondering what goes wrong there :D But it's not a problem. git will fix this when merging into main.

shredEngineer:
> Have you cherry picked my change? :D Looks like it. I am really wondering what goes wrong there :D But it's not a problem. git will fix this when merging into main.

Not knowingly. I should really do some kind of git course; I can't always bug you with my incompetence regarding this issue. :D

src/book.py
provinzio: I thought you switch the if-case with the calculation. Like so. Less code. Does the same. :D

```
                    if dup_state[_type] == dup_skip[_type]:
                        continue
                    dup_state[_type] = (dup_state[_type] + 1) % 2
```

provinzio: But my thought is stupid XD

src/price_data.py
```diff
@@ -124,13 +124,22 @@ def _get_price_kraken(self, base_asset: str, utc_time: datetime.datetime, quote_
...
-        If the 10 minutes-offset is already too large, recursively retry by reducing the offset step, down to 1 minute.
+        If the initial offset is already too large, recursively retry by reducing the offset step, down to 1 minute.
 
         Documentation: https://www.kraken.com/features/api
+
+        Args:
+            base_asset (str): Base asset
+            utc_time (datetime.datetime): Target time (time of the trade)
+            quote_asset (str): Quote asset
+            minutes_step (int): Initial time offset for consecutive Kraken API requests
```
provinzio, line 139: [...] .Defaults to 10.

Each docstring line should end with an `.`.

provinzio (approved):
Hey @shredEngineer,

I only have another tiny comment. Otherwhise this looks awesome, as I said before. I appreciate the time you put into this.

Can you rebase your commits together? (Take care, that you don't merge my commit with yours together.)

So far, the code is ready to be merged into main.

src/book.py
shredEngineer: No problem! :D

shredEngineer:
> I only have another tiny comment. Otherwhise this looks awesome, as I said before. I appreciate the time you put into this.

Thank **you** for putting everything else together! :)

> Can you rebase your commits together? (Take care, that you don't merge my commit with yours together. While rebasing you could remove the cherry picked commit from your branch.)

Alright, I think I may have managed to do so correctly this time!

Issue #21, opened 2021-03-06
Kraken API requests frequency exceeds standard rate limit


wearymanateevedaknotstabooresisting:
The current request frequency has a too small period, resulting in the excess of the Kraken specified rate limit.

This leads to non-requested _get_price_kraken calls, which return 0 and seem to not be re-requested, thus leading to false calculations afterwards (an may be won't be re-queried in another run, since 0 is saved into the DB? Need confirmation, here).

```warning
2021-03-06 13:38:05,744 price_data   WARNING  Querying trades for #PAIR# at YYYY-mm-dd HH-MM-SS+T (offset=20m): Could not retrieve trades: ['EAPI:Rate limit exceeded']
```

provinzio:
Oh, I thought `raise_for_status` ([here](https://github.com/provinzio/CoinTaxman/blob/main/src/price_data.py#L160)) would catch the teapot/rate limit.

It looks like you are right.

#16 could fix this, if it's implemented for different exchanges (which would include Kraken)

wearymanateevedaknotstabooresisting:
> Calling the public endpoints at a frequency of 1 per second (or less) would remain within the rate limits, but exceeding this frequency could cause the calls to be rate limited. If the rate limits are reached, additional calls would be restricted for a few seconds (or possibly longer if calls continue to be made while the rate limits are active).

-Kraken, [https://support.kraken.com/hc/en-us/articles/206548367-What-are-the-API-rate-limits-#1]

A hot fix would be to increase the random uniform `delay` to a range between [1,2]

wearymanateevedaknotstabooresisting:
> Oh, I thought raise_for_status (here) would catch the teapot/rate limit.
- @provinzio 

Actually, Kraken does still send a status code in the 800s if the rate is limited. `raise_for_status` is only triggered with 400/500s codes for server/client errors (https://requests.readthedocs.io/en/master/user/quickstart/). Therefore, no exception will be raised.

provinzio:
> 
> 
> > Oh, I thought raise_for_status (here) would catch the teapot/rate limit.
> > -@provinzio
> 
> Actually, Kraken does still send a status code in the 800s if the rate is limited. `raise_for_status` is only triggered with 400/500s codes for server/client errors (https://requests.readthedocs.io/en/master/user/quickstart/). Therefore, no exception will be raised.
> 
[...]

PR #22, opened 2021-03-06
Resolve #18 to support older Kraken account statements than October 2020

### LINKED ISSUE #18: Support old Kraken statement files
The Kraken ledger file changed over the years. While current Kraken users may be able to generate a fresh statement (in the supported format), we could adjust our code, to support old files.

wearymanateevedaknotstabooresisting:
Resolves #18

Addresses #18, please refer to the issue problem description.
Addresses #21 with a hot fix for avoiding Kraken API rate limiting.

# Changes

Classifying account statements of exchanges by mapping columns of Kraken export files now supports old Kraken files from September 2020 and before. The additional read method has been linked to the original one, with the process being not different. The line where the `subtype` was used has been changed to neglect it. It hasn't been used for functionality either way.

The `""` column bug has been resolved by removing the erroneous column name.

# Implementation

- A new exchange mapping `kraken_ledgers_old` has been added, with no `subtype`.
- The empty `""` kraken trade column name has been removed. It does not exist in any Kraken export file.
- The read method `_read_kraken_ledgers_old` has been added, linking to `_read_kraken_ledgers` which is no different.
- The `subtype` has been removed from `_read_kraken_ledgers`, not contributing anything (yet) either way.
  - This should be discussed, if `subtype` is of use with respect to a special tax calculation law/process.
- The account statement headers are now compared, again, in the parsing function to prevent duplication of huge amounts of code or heavily re-write de parsing.
- The Kraken API requests are now retried with an exponential sleeping time
- Updated `requirements.txt` via `pip freeze -r requirements.txt > requirements.txt` and deleting the header comment.
- Updated `.gitignore` to account for VSCode IDE and Python virtual environment directories.


wearymanateevedaknotstabooresisting:
@provinzio how do I pre-run CodeQL locally, to verify my changes in before the server CI check?

wearymanateevedaknotstabooresisting:
@provinzio @shredEngineer please note that I happily invite you to directly contribute to PRs and code I wrote, by directly committing to the pull requests - I added you to my fork. Especially with a limited number of contributors, projects move faster this way.

.gitignore
```diff
@@ -10,4 +10,10 @@ export/*
 # Development folder
 dev/
 
+# IDE
+.vscode/
```
provinzio, line 14: If we remove .vscode from git, we should delete the current .vscode folder.

requirements.txt
```diff
@@ -1,2 +1,7 @@
+certifi==2020.12.5
+chardet==4.0.0
+idna==2.10
 python-dateutil==2.8.1
 requests==2.25.1
+six==1.15.0
+urllib3==1.26.3
```
provinzio, line 7: Why should we list these modules? Can't we leave the dependancies to the imported modules?

provinzio:
> @provinzio how do I pre-run CodeQL locally, to verify my changes in before the server CI check?

I believe, you use VSC? It should be able by adding the corresponding extension and add the GitHub database. 

But adding CodeQL to the repo was more like playing with GitHub workflows. I am considering to remove it, because I do not see a real reason to stay with it.

.gitignore
wearymanateevedaknotstabooresisting: Actually, I didn't notice that .vscode directory was already part of the repository but thought I introduced it through my VSCode development. However, if the pre-settings are not really necessary, I'd drop that folder but keep the ignore line. 

As you wish.

requirements.txt
wearymanateevedaknotstabooresisting: It has been best practice to specify all dependencies and their dependencies in the most explicit manner possible. This gives a really well defined development environment and prevents incompatibilities / inconsistencies between dev setups.

However, I'm adaptive, if you like to maintain with minimal requirements / most flexibility.

wearymanateevedaknotstabooresisting:
> But adding CodeQL to the repo was more like playing with GitHub workflows. I am considering to remove it, because I do not see a real reason to stay with it.

In that case I won't bother to use it myself. Therefore, I am pro towards leaving it out, except unittest / pytest is introduced and runs flawlessly with CodeQL (haven't used that one myself, only worked with GitLab-CI in my professional environment.

.gitignore
provinzio: I mean. I like the vscode file in their. It reduces my setup time on different machines. But it's not actually a part of CoinTaxman. I though when you remove this, it might be a good idea in general to keep the repo clean

provinzio:
I removed the CodeQL workflow for now (19c45bc956227f27e0ed02bf365748e4563a3d8a).

.gitignore
provinzio: I removed the .vscode Folder (31f172419de955e707fbf340f94b3d000e3c019d).

provinzio:
Good work so far! Does this PR completely solve #18 or are there older formats you know about? (You can use "Closes/Resolves #xxx" in the PR to automatically link the issue with the PR, if it is supposed to solve it directly. Have a look [here](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for further information.

wearymanateevedaknotstabooresisting:
I tested this PR with old and new account statements from Kraken export for the same time interval. The runtime succeeded with identical results. Further, the Kraken API request limitation issue is hot fixed with a work-around of retrying requests with exponentially growing delay.

However, this PR does not proof for the correctness of the general tax calculation. It was a pure runtime fix of #18.

@provinzio and @shredEngineer: The PR is now merge ready in my opinion. Please review and test if you have account statements at hand.

provinzio (approved):
I reviewed your code and just made a slight adjustment (https://github.com/provinzio/CoinTaxman/pull/22/commits/f1bc59707140d348b7a28a0984ba6e23a95d6e6a).

@shredEngineer could you run the tests with your account statements? I am not a Kraken user.

provinzio:
> However, this PR does not proof for the correctness of the general tax calculation. It was a pure runtime fix of #18.

I am a bit confused, I thought this resolves #18 (PR description).

So it should only adress #18 and close #21?

shredEngineer:
> @shredEngineer could you run the tests with your account statements? I am not a Kraken user.

Works well for me! But note that I don't have account statements reaching back several years or containing "rare" asset pairs, so my testing doesn't seem to cover the edge cases which @wearymanateevedaknotstabooresisting encountered.

wearymanateevedaknotstabooresisting:
> I am a bit confused, I thought this resolves #18 (PR description).

> So it should only adress #18 and close #21?

This PR fixes the issues I had with my older Kraken account statements, so that the outcome is the same as with the newer account statement format. So it resolves #18 regarding parsing of old files.

However, I did not check the calculation based on retrieved prices and the Kraken FIFO calculation in general which was not scope of #18. This is still a to-do for me, but will need another issue and new PR if there are errors.

Regarding #21, this PR can be seen as hot-fix to make it work at all with the current implementation how to request the API. But in #21 we talked about more, like general behavior of the program when prices could not be retrieved at all and user action/decision is demanded. This is not scope of this PR.

From my point of view, this PR resolved its scope.

provinzio:
Ok, I just asked for clarification. That's also my point of view. Thank you for finding and fixing the bug.

wearymanateevedaknotstabooresisting:
Thanks for merging and the opportunity to contribute.

Issue #29, opened 2021-03-16
New exchange: Coinbase Pro 


sibbl:
I'd love to see Coinbase Pro as part of this tool and already started implementing it 🚀

For getting the prices, the [official API for "Getting historical rates"](https://docs.pro.coinbase.com/#get-historic-rates) can be used. My plan would be to get the nearest historic rate with the smallest granularity and use the average of `high` and `low` values.

However, there's the problem that Coinbase doesn't support XRP anymore and has removed the historical data from their API. I'm not sure how to tackle this. Should the tool be extended to use other exchances as kind of a fallback strategy?

provinzio:
We also thought about a fallback in #15. We can't get around it. In tax calculation it's no option to miss a price for a transaction.

You might want to have a look at #16. We started to implement `ccxt` which supports numerous exchanges. The first approach covers only binance though, but we planned to spread it to other exchanges which could include Coinbase Pro (it seems to be supported by `ccxt`, too).

It might be possible, that you assist @scientes on his PR.

PR #31, opened 2021-03-19
Add Coinbase Pro and extend Coinbase support

sibbl:
The data of Coinbase Pro is supposed to be generated under [Profile > Statements](https://pro.coinbase.com/profile/statements) by clicking "Generate" > ~"Account"~ "Fills" and selecting CSV as the format.

The only remaining limitation is that Coinbase Pro API doesn't return data for XRP anymore, as discussed in #29 already.

Besides adding Coinbase Pro, there are some other minor changes in this PR as well.

1. I added `Receive` and `Coinbase Earn` operation mappings for Coinbase.
2. I also added `Convert` support for Coinbase, it was required to get the rates from somewhere. For consistency I've decided to never use the rates directly from the Coinbase CSV anymore and always get the pricing information from the Coinbase Pro API.
3. To avoid errors with my data, I needed to also check `base_asset` being `BTC` to swap the coins to get the correct pricing. I also applied this to the get price method of Binance.

I'd also suggest to update the Wiki for how to use both Coinbase and Coinbase Pro when this PR will be done.

I'm open to honest feedback as Python is not (yet?) my go to language 😊

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -128,7 +128,7 @@ class Book:
         operation_mapping = {
             "Receive": "Deposit",
             "Send": "Withdraw",
-            "Coinbase Earn": "Buy"
+            "Coinbase Earn": "Buy",
         }
 
         with open(file_path, encoding="utf8") as f:
@@ -197,16 +197,24 @@ class Book:
 
                 if operation == "Convert":
                     # parse change + coin from remark which is in format "0,123 ETH to 0,456 BTC"
-                    remark_target = remark.split(' to ')[-1]
-                    remark_target_parts = remark_target.split(' ')
-                    _convert_change = remark_target_parts[0].replace(',', '.')
+                    remark_target = remark.split(" to ")[-1]
+                    remark_target_parts = remark_target.split(" ")
+                    _convert_change = remark_target_parts[0].replace(",", ".")
                     convert_change = misc.xdecimal(_convert_change)
                     convert_coin = remark_target_parts[1]
 
-                    self.append_operation("Sell", utc_time, platform,
-                                          change, coin, row, file_path)
-                    self.append_operation("Buy", utc_time, platform,
-                                          convert_change, convert_coin, row, file_path)
+                    self.append_operation(
+                        "Sell", utc_time, platform, change, coin, row, file_path
+                    )
+                    self.append_operation(
+                        "Buy",
+                        utc_time,
+                        platform,
+                        convert_change,
+                        convert_coin,
+                        row,
+                        file_path,
+                    )
                 else:
                     self.append_operation(
                         operation, utc_time, platform, change, coin, row, file_path
@@ -215,12 +223,24 @@ class Book:
                     if operation == "Sell":
                         assert isinstance(eur_subtotal, decimal.Decimal)
                         self.append_operation(
-                            "Buy", utc_time, platform, eur_subtotal, "EUR", row, file_path
+                            "Buy",
+                            utc_time,
+                            platform,
+                            eur_subtotal,
+                            "EUR",
+                            row,
+                            file_path,
                         )
                     elif operation == "Buy":
                         assert isinstance(eur_subtotal, decimal.Decimal)
                         self.append_operation(
-                            "Sell", utc_time, platform, eur_subtotal, "EUR", row, file_path
+                            "Sell",
+                            utc_time,
+                            platform,
+                            eur_subtotal,
+                            "EUR",
+                            row,
+                            file_path,
                         )
 
                     if eur_fee:
@@ -241,12 +261,25 @@ class Book:
             # Skip header.
             next(reader)
 
-            for portfolio, trade_id, product, operation, _utc_time, _size, size_unit, _price, _fee, total, price_fee_total_unit in reader:
+            for (
+                portfolio,
+                trade_id,
+                product,
+                operation,
+                _utc_time,
+                _size,
+                size_unit,
+                _price,
+                _fee,
+                total,
+                price_fee_total_unit,
+            ) in reader:
                 row = reader.line_num
 
                 # Parse data.
                 utc_time = datetime.datetime.strptime(
-                    _utc_time, "%Y-%m-%dT%H:%M:%S.%fZ")
+                    _utc_time, "%Y-%m-%dT%H:%M:%S.%fZ"
+                )
                 utc_time = utc_time.replace(tzinfo=datetime.timezone.utc)
                 operation = operation_mapping.get(operation, operation)
                 size = misc.xdecimal(_size)
@@ -266,18 +299,40 @@ class Book:
                 assert size_unit
                 assert price_fee_total_unit
 
-                self.append_operation(operation, utc_time, platform,
-                                      size, size_unit, row, file_path)
+                self.append_operation(
+                    operation, utc_time, platform, size, size_unit, row, file_path
+                )
 
                 if operation == "Sell":
-                    self.append_operation("Buy", utc_time, platform,
-                                          total_price, price_fee_total_unit, row, file_path)
+                    self.append_operation(
+                        "Buy",
+                        utc_time,
+                        platform,
+                        total_price,
+                        price_fee_total_unit,
+                        row,
+                        file_path,
+                    )
                 elif operation == "Buy":
-                    self.append_operation("Sell", utc_time, platform,
-                                          total_price, price_fee_total_unit, row, file_path)
+                    self.append_operation(
+                        "Sell",
+                        utc_time,
+                        platform,
+                        total_price,
+                        price_fee_total_unit,
+                        row,
+                        file_path,
+                    )
                 if fee:
-                    self.append_operation("Fee", utc_time, platform,
-                                          fee, price_fee_total_unit, row, file_path)
+                    self.append_operation(
+                        "Fee",
+                        utc_time,
+                        platform,
+                        fee,
+                        price_fee_total_unit,
+                        row,
+                        file_path,
+                    )
 
     def _read_kraken_trades(self, file_path: Path) -> None:
         log.error(
--- a/src/book.py
+++ b/src/book.py
@@ -196,11 +196,12 @@ class Book:
                 assert change
 
                 if operation == "Convert":
-                    # parse change + coin from remark which is in format "0,123 ETH to 0,456 BTC"
+                    # parse change + coin from remark which is
+                    # in format "0,123 ETH to 0,456 BTC"
                     remark_target = remark.split(" to ")[-1]
                     remark_target_parts = remark_target.split(" ")
                     _convert_change = remark_target_parts[0].replace(",", ".")
-                    convert_change = misc.xdecimal(_convert_change)
+                    convert_change = misc.force_decimal(_convert_change)
                     convert_coin = remark_target_parts[1]
 
                     self.append_operation(
@@ -282,8 +283,8 @@ class Book:
                 )
                 utc_time = utc_time.replace(tzinfo=datetime.timezone.utc)
                 operation = operation_mapping.get(operation, operation)
-                size = misc.xdecimal(_size)
-                price = misc.xdecimal(_price)
+                size = misc.force_decimal(_size)
+                price = misc.force_decimal(_price)
                 fee = misc.xdecimal(_fee)
                 total_price = size * price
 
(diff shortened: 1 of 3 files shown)


src/book.py
```diff
@@ -188,36 +188,95 @@ def _read_coinbase(self, file_path: Path) -> None:
 
                 # Unused variables.
                 del eur_total
-                del remark
```
provinzio, line 191: To keep the current format. Rename `_eur_spot` to `eur_spot` and add `del eur_spot`.

src/book.py
```diff
@@ -188,36 +188,95 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+                        self.append_operation(
+                            "Buy", utc_time, platform, eur_subtotal, "EUR", row, file_path
+                        )
+                    elif operation == "Buy":
+                        assert isinstance(eur_subtotal, decimal.Decimal)
+                        self.append_operation(
+                            "Sell", utc_time, platform, eur_subtotal, "EUR", row, file_path
+                        )
+
+                    if eur_fee:
```
provinzio, line 225: Is  this supposed to be indented?
Are `Convert` operations always without fees?

src/book.py
```diff
@@ -188,36 +188,95 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+                row = reader.line_num
+
+                # Parse data.
+                utc_time = datetime.datetime.strptime(
+                    _utc_time, "%Y-%m-%dT%H:%M:%S.%fZ")
+                utc_time = utc_time.replace(tzinfo=datetime.timezone.utc)
+                operation = operation_mapping.get(operation, operation)
+                size = misc.xdecimal(_size)
+                price = misc.xdecimal(_price)
+                fee = misc.xdecimal(_fee)
```
provinzio, line 253: Consider using `misc.force_decimal`, when the value is supposed to be a `Decimal`.

provinzio: Ah, I see.. I believe you are not using `mypy`. Pulling the rebasing to origin/main should add your branch our new linter workflow which helps you to detect errors like this and more.

For easy local usage, I created a `Makefile` with the relevant commands. If you have not `make` installed, you can copy/paste them into the terminal.

src/price_data.py
```diff
@@ -97,7 +97,7 @@ def _get_price_binance(
             and data.get("code") == -1121
             and data.get("msg") == "Invalid symbol."
         ):
-            if quote_asset == "BTC":
+            if quote_asset == "BTC" or base_asset == "BTC":
```
provinzio, line 100: `if "BTC" in (base_asset, quote_asset)`

provinzio, line 100: When does the case base_asset == "BTC" happen?

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+
+            btc = self.get_price("coinbase_pro", base_asset, utc_time, "BTC")
+            quote = self.get_price("coinbase_pro", "BTC", utc_time, quote_asset)
+            return btc * quote
+
+        response.raise_for_status()
+
+        if len(data) == 0:
+            log.warning("Coinbase Pro offers no price for `%s` at %s", symbol, utc_time)
+            return 0
```
provinzio, line 202: This function is supposed to return a decimal. Return `decimal.Decimal()`.

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+    def _get_price_coinbase(self, base_asset: str, utc_time: datetime.datetime, quote_asset: str, swapped_symbols: bool = False) -> decimal.Decimal:
+        # Use Coinbase Pro prices
+        return self._get_price_coinbase_pro(base_asset, utc_time, quote_asset, swapped_symbols)
+
+    @misc.delayed
+    def _get_price_coinbase_pro(self, base_asset: str, utc_time: datetime.datetime, quote_asset: str, swapped_symbols: bool = False) -> decimal.Decimal:
+        """Retrieve price from Coinbase Pro official REST API.
+
+        The price is calculated as the average price of the opening and closing price
+        around 1 minute around `utc_time`.
```
provinzio, line 146: remove first around

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+        Args:
+            start	        Start time in ISO 8601
+            end	            End time in ISO 8601
+            granularity	    Desired timeslice in seconds (60, 300, 900, 3600, 21600 or 86400)
+
+        Raises:
+            RuntimeError: Unable to retrieve price data.
+
+        Returns:
+            Decimal: Price of asset pair.
```
provinzio, line 162: decimal.Decimal

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+
+        The price is calculated as the average price of the opening and closing price
+        around 1 minute around `utc_time`.
+
+        None existing pairs like `TWTEUR` are calculated as
+        `TWTBTC * BTCEUR`.
+
+        Documentation: https://docs.pro.coinbase.com
+
+        Args:
```
provinzio, line 153: This tag is only for function parameters and I wouldn't place the API description in the function docstring. Just place it in the code.

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+        url = root_url + sub_url
+
+        log.debug("Calling %s", url)
+        response = requests.get(url)
+        data = json.loads(response.text)
+
+        # Some combinations do not exist (e.g. `TWTEUR`), but almost anything
+        # is paired with BTC. Calculate `TWTEUR` as `TWTBTC * BTCEUR`.
+        if isinstance(data, dict) and data.get("message") == "NotFound":
+            if quote_asset == "BTC" or base_asset == "BTC":
```
provinzio, line 179: It might be cleaner, if we move this to a separate function, since we use the exact same stuff again.

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+        # Find closest timestamp match
+        target_timestamp = misc.to_ms_timestamp(utc_time)
+        data_timestamps_ms = [int(float(d[2]) * 1000) for d in data]
+        closest_match_index = (
+            bisect.bisect_left(data_timestamps_ms, target_timestamp) - 1
+        )
+
+        # Use item in the middle if no closest match found
+        if closest_match_index == -1:
+            closest_match_index = int(len(data) / 2)
```
provinzio, line 213: `len(data) // 2` does the same.

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+
+        # Find closest timestamp match
+        target_timestamp = misc.to_ms_timestamp(utc_time)
+        data_timestamps_ms = [int(float(d[2]) * 1000) for d in data]
+        closest_match_index = (
+            bisect.bisect_left(data_timestamps_ms, target_timestamp) - 1
+        )
+
+        # Use item in the middle if no closest match found
+        if closest_match_index == -1:
```
provinzio, line 212: The searched timestamp is older than all other timestamps when the function returns -1... It doesn't make sense to take the middle value.

src/price_data.py
```diff
@@ -133,6 +133,91 @@ def _get_price_binance(
...
+            RuntimeError: Unable to retrieve price data.
+
+        Returns:
+            Decimal: Price of asset pair.
+        """
+
+        root_url = "https://api.pro.coinbase.com"
+        symbol = f"{base_asset}-{quote_asset}"
+        startTime = misc.to_iso_timestamp(utc_time - datetime.timedelta(minutes=5))
+        endTime = misc.to_iso_timestamp(utc_time + datetime.timedelta(minutes=5))
```
provinzio, line 168: We should catch the case, when the offset window is to shallow and might want to increase it a little bit. Have a look at `_get_price_kraken` for that.

provinzio (changes requested):
Hey @sibbl, thank you for your time you put into this! Could you provide me with an example account statment for further testing?

src/book.py
sibbl: There are no direct fees as Coinbase works with spread margins up to 2%: https://help.coinbase.com/en/coinbase/trading-and-funding/cryptocurrency-trading-pairs/convert-cryptocurrency-faq 

However I'll try to do some calculations again and check whether it's correct what I'm doing here. Especially since we got some information about the used prices in the CSV and have access to the Coinbase Pro price information. Perhaps a hybrid approach is necessary here.

sibbl: I've updated the Coinbase pricing mechanism by using the spot prices from the CSV again. See this sample:

```
Timestamp: 2019-04-03T08:51:36Z
Transaction Type: Convert
Asset: BTC
Quantity Transacted: 0.045368
EUR Spot Price at Transaction: 4396.49
EUR Subtotal: 384.43
EUR Total (inclusive of fees): 199.46
EUR Fees: 184.97
Notes: "Converted 0,045368 BTC to 0,8654877 BCH"
```

In our code this would be:
```python
coin = "BTC"
change = Decimal("0.045368")
convert_coint = "BCH"
convert_change = "0.8654877"
```

Which means that I can calculate the exchange rate of BCH by calculating `eur_total / convert_change` which is `199,45995832 EUR / 0,8654877 BCH = 230,459610598741 EUR/BCH`

A sanity check using the Coinbase Pro API: the BCH price at 2019-04-03 08:51:36 was 227.815 EUR/BCH

So although there are no fees, it's indeed important to use the values from the CSV as even Coinbase and Coinbase Pro prices differ and this is how they can allow converting without additionally charging fees.

Just out of curiosity because luckily it's irrelevant for the algorithm: if somebody finds out what the values of Coinbase's columns "EUR Subtotal" and "EUR Fees" are, please let me know. I know for sure that I didn't pay 184.97 EUR of fees 😊

jhoogstraat:
The correct statement to generate is called "Fills" instead of "Account".

src/book.py
sibbl: This should probably be fine since now. I've just configured VS Code properly:

```json
{
    "python.linting.enabled": true,
    "python.linting.flake8Enabled": true,
    "python.linting.mypyEnabled": true
}
```

sibbl:
> The correct statement to generate is called "Fills" instead of "Account".

Thanks for your hint, that's correct. I've update my main comment to avoid further confusion right away.
Should do this kind of documentation earlier next time 👍

src/price_data.py
sibbl: I should've investigated this right away as I cannot reproduce it anymore. 

I can imagine it was some kind of issue I had in combination with some local changes where I used Binance as a fallback for Coinbase pairs (e.g. because XRP is not part of the Coinbase API anymore).

I've reverted the change.

src/price_data.py
sibbl: I've updated the implementation, which is now inspired by Kraken.

src/price_data.py
sibbl: It seems that these lines are not necessary anymore since I don't use the API for Coinbase (non-Pro) anymore - so they're removed now.

src/price_data.py
sibbl: I'm not taking the middle value anymore, now that the implementation increases the time window if bisect couldn't find some item before and after the timestamp.

src/book.py
```diff
@@ -188,36 +188,95 @@ def _read_coinbase(self, file_path: Path) -> None:
...
-                self.append_operation(
-                    operation, utc_time, platform, change, coin, row, file_path
-                )
+                if operation == "Convert":
+                    # parse change + coin from remark which is in format "0,123 ETH to 0,456 BTC"
+                    remark_target = remark.split(' to ')[-1]
+                    remark_target_parts = remark_target.split(' ')
+                    _convert_change = remark_target_parts[0].replace(',', '.')
+                    convert_change = misc.xdecimal(_convert_change)
+                    convert_coin = remark_target_parts[1]
```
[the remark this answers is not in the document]
sibbl: The regex is implemented now. I initially decided against it because I mostly don't think regular expressions are the most readable and maintainable way but I think it turned out quite okay 😊

provinzio (approved):
Hey @sibbl,

your changes completly fulfill the scope of the issue. I really appreciate the time you've put into this and believe that many users will do so, too.

I won't bother you with my small change (see below) and will adjust it myself, as you already did a splendid job.

PR #37, opened 2021-04-19
Balance queue: fix LIFO queue and not_sold assertion

sibbl:
* The LIFO queue code seems to be outdated, so I fixed the inheritance and adapted the `_get` method.
* The `not_sold > 0`  assertion failed because it was exactly equal to `0`. I think this should also be fine.

### DIFF
--- a/src/balance_queue.py
+++ b/src/balance_queue.py
@@ -124,7 +124,7 @@ class BalanceQueue:
                 return None
 
             not_sold = bop.op.change - bop.sold
-            assert not_sold >= 0
+            assert not_sold > 0
 
             if not_sold > change:
                 bop.sold += change
--- a/src/balance_queue.py
+++ b/src/balance_queue.py
@@ -18,7 +18,6 @@ import collections
 import dataclasses
 import decimal
 import logging
-import queue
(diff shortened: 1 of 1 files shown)


src/balance_queue.py
```diff
@@ -125,7 +124,7 @@ def sell(self, change: decimal.Decimal) -> Optional[list[transaction.SoldCoin]]:
                 return None
 
             not_sold = bop.op.change - bop.sold
-            assert not_sold > 0
+            assert not_sold >= 0
```
provinzio, line 127: bop should have been removed from the queue, if it has no coins left which can be sold. So this should be > 0. Otherwise we made a mistake earlier.

provinzio (changes requested):

src/balance_queue.py
sibbl: I reverted the commit which added this change. Will try to find some time and check why this happens on my side then. 👍

provinzio:
I cherry-picked the LIFO fix. Thanks for your contribution. I am looking forward to your further findings.

provinzio:
FYI I might have found the culprit of the faulty assertion.

Issue #19, opened 2021-03-06
Offer tests account statements for code test of exchange integrations


wearymanateevedaknotstabooresisting:
Hereby, I request to offer artificial account statements (best, real statements with altered data), to test exchange integrations regarding valid implementation.

This includes the by-hand calculation of taxes with respect to the applicable law which is supported.

It should be noted, that all supported tax laws should be tested with each integration. Best, in a combined test file, which is the real world use-case when exporting exchange activity by users.

shredEngineer:
I don't have by-hand calculation of taxes, but here are my test account statements for Kraken (new format), which I also uploaded in some other issue here. Maybe this is useful to someone.
[kraken-example-ledgers.csv.txt](https://github.com/provinzio/CoinTaxman/files/6100449/kraken-example-ledgers.csv.txt)
[kraken-example-trades.csv.txt](https://github.com/provinzio/CoinTaxman/files/6100450/kraken-example-trades.csv.txt)

wearymanateevedaknotstabooresisting:
Thanks, that provides a bare minimum test. I guess we'd need one for every exchange (times) tax law application

sibbl:
I'd also like to share some data as I found some limitations for Coinbase and Binance with it already. I'm still trying to get at least one end-to-end run finish successfully before I can contribute though...

Would it also make sense to think about writing unit tests using this data?

provinzio:
It would be awesome to provide unit testing.

PR #40, opened 2021-04-27
Test account statements for unittest

scientes:
This should be a starting point for #19 
addtions welcome


scientes:
this should be a starting point

provinzio:
Hey @scientes,
we should add some test functions before we merge the test files.

PR #47, opened 2021-05-08
Add sell price to TaxEvent and thus also CSV export

sibbl:
This PR adds two new columns to the exported CSV:
* ~Buy Price in {FIAT}~
* Sell Price in {FIAT}

The reason was that I needed to add these values to my taxation software and it wasn't possible to calculate either from the data already available in the CSV.

~The data is a bit redundant because the "taxed gain" is the difference between both new columns, so only one new column would be required to do your own maths in the spreadsheet application of your choice. However I thought it would be helpful to have both and avoid as much content transformation as possible.~

Let me know what you think of this change and whether it's okay to "break" the existing format. Perhaps it would make sense to add the amount of fees as the "Sell price" and leave "Buy price" as 0, but it might also confuse people.

In the long term, we might also think about having different formats which can be used for exporting. E.g. CoinTracking has its own format, which is supported by WISO software (see https://cointracking.info/tax/tax_details/WISO.php?tax=foobar ) - probably other taxation softwares already have or will have something similar in the future.

Update 2021-05-10: initially the "buy price" was also part of this PR. I removed it and also updated this comment to avoid further confusion.

### DIFF
--- a/src/taxman.py
+++ b/src/taxman.py
@@ -110,7 +110,7 @@ class Taxman:
                 if self.in_tax_year(op) and coin != config.FIAT:
                     taxation_type = "Sonstige Einkünfte"
                     # Price of the sell.
-                    total_win = self.price_data.get_cost(op)
+                    sell_price = self.price_data.get_cost(op)
                     taxed_gain = decimal.Decimal()
                     # Coins which are older than (in this case) one year or
                     # which come from an Airdrop, CoinLend or Commission (in an
@@ -130,8 +130,9 @@ class Taxman:
                             )
                             and not sc.op.coin == config.FIAT
                         ):
-                            partial_win = (sc.sold / op.change) * total_win
-                            taxed_gain += partial_win - self.price_data.get_cost(sc)
+                            partial_sell_price = (sc.sold / op.change) * sell_price
+                            sold_coin_cost = self.price_data.get_cost(sc)
+                            taxed_gain += partial_sell_price - sold_coin_cost
                     remark = ", ".join(
                         f"{sc.sold} from {sc.op.utc_time} "
                         f"({sc.op.__class__.__name__})"
@@ -141,8 +142,7 @@ class Taxman:
                         taxation_type,
                         taxed_gain,
                         op,
-                        total_win,
-                        total_win - taxed_gain,
+                        sell_price,
                         remark,
                     )
                     self.tax_events.append(tx)
@@ -242,7 +242,6 @@ class Taxman:
                 "Action",
                 "Amount",
                 "Asset",
-                f"Buy Price in {config.FIAT}",
                 f"Sell Price in {config.FIAT}",
                 "Remark",
             ]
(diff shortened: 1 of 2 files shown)


src/taxman.py
```diff
@@ -137,7 +137,14 @@ def _evaluate_taxation_GERMANY(
                         f"({sc.op.__class__.__name__})"
                         for sc in sold_coins
                     )
-                    tx = transaction.TaxEvent(taxation_type, taxed_gain, op, remark)
+                    tx = transaction.TaxEvent(
+                        taxation_type,
+                        taxed_gain,
+                        op,
+                        total_win,
+                        total_win - taxed_gain,
```
provinzio, line 145: Might be enough to save only `total_win` in `TaxEvent` as we can calculate this on the fly.

src/taxman.py
```diff
@@ -137,7 +137,14 @@ def _evaluate_taxation_GERMANY(
                         f"({sc.op.__class__.__name__})"
                         for sc in sold_coins
                     )
-                    tx = transaction.TaxEvent(taxation_type, taxed_gain, op, remark)
+                    tx = transaction.TaxEvent(
+                        taxation_type,
+                        taxed_gain,
+                        op,
+                        total_win,
```
provinzio, line 144: You swapped sell and buy price. `total_win` is the sell_price and should come second.

We might want to rename `total_win` to avoid further confusion.

provinzio (changes requested):
If we leave it like that, we should make clear, that the buy price comes from a really weird calculation in the background.(e.g.
- selling only StakingInterests. 0 buy price
- selling multiple buy positions: average buy price
- selling StakingInterest and buy positions: some weird "virtual" buy price
- selling buy positions which are older than 1 year (non-taxable) and younger than 1 year: also resulting in a weird "virtual" buy price

I don't mind to add the sell price to the csv. If you need the "buy" price.. well, we can add it, but should label it "virtual" buy price or something like that and add documentation what it means. 

Currently I am missing a good idea how we could add documentation to the csv file.
Might be easier to avoid adding "virtual" buy price to the csv.

sibbl:
I'd suggest to only include the "sell price" then, as other relevant calculations for the tax may be done with this and the taxed gain column? I think it's easier than exposing the calculated value as "buy price" and explaining it.

I just reverted the changes accordingly and renamed `total_win` to `sell_price`.

Let me know if I should open another PR or if squashing is okay to avoid the (kind of) revert commit to appear in the commit history.

src/taxman.py
sibbl: The swapping should be irrelevant now that there's only the "sell price" in the CSV. Also, I renamed `total_win` to `sell_price` or would you suggest another naming? :)

src/taxman.py
sibbl: Should have become irrelevant now due to removal of buy price column.

provinzio:
Squashing should be good enough.

I like your idea of adding different exports for different needs. Especially when the format is supported by a (major) taxation software. Over all, I don't mind to add multiple export formats, as long as they server a purpose. Feel free to open an Issue or PR.

Issue #59, opened 2021-08-15
Implement Binance Tax reporting API


scientes:
https://www.binance.com/en/support/announcement/57d68e2a247744aa980ab82f6708522d

sebastianelsner:
The tax api is only the normal api just with specific read-only rights. Does not seem too much of a boon for me.

scientes:
ok good to know i just knew that it existed, but for some things in binance like etf staking and liquid swap and such  so we do not get all needed data just through the csv files, this could be fixed through the api

PR #65, opened 2021-08-16
add liquid swap

### LINKED ISSUE #51: [Feature Request] Add DeFi options
source for german tax law regarding defi: Like coin lending but with a small twist
https://winheller.com/blog/decentralized-finance-defi-lending-steuern/

scientes:
closes #51 
does not add support for defi lending  only for liquid swap this should maybe be another issue

also currently binance classifies the rewards as "distribution" and we therefore as airdrop which im not quite sure is correct, the same goes for eth2 staking

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -78,6 +78,8 @@ class Book:
             "Launchpool Interest": "StakingInterest",
             "Cash Voucher distribution": "Airdrop",
             "Super BNB Mining": "StakingInterest",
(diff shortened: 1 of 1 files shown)


scientes:
This may have not fixed this entirely, because binance does not deliver the distribution histroy for swaps in the transaction log, this can be maybe be fixed with #59 or by querying the api with an api key

PR #72, opened 2021-09-07
Add support for containerizing CoinTaxman

jhoogstraat:
This pr adds support for containerizing this project by adding a dockerfile and rules to the Makefile.
To be able to set the country and tax year, they can now be set via environment variables.


Makefile
```diff
@@ -33,11 +45,15 @@ cleanrun: clean run
 # Install requirements
 install:
 	python -m pip install --upgrade pip
-	pip install -r requirements.txt -r requirements-dev.txt
+	pip install -r requirements.txt
+
+install-dev:
```
provinzio, line 50: Install dev is missing the regular install. We could write it like this to extend install with Installation of dev files. 

```
install_dev: install
  Pip install req-dev....
````

**Review by provinzio — CHANGES_REQUESTED** [review]

**Comment by provinzio** [comment]
Please run isort (e.g. make isort) to fix the import order.

File: Makefile (hunk 1 above)
Comment (jhoogstraat) [reply Makefile:50]: Good idea.
Do you prefer underscore (install_dev instead of install-dev) ?
To be consistent, I'll change run-container too.

Comment (provinzio) [reply Makefile:50]: My fault.  Let's stick with "install-dev".

**Comment by jhoogstraat** [comment]
Maybe add a make rule to push an image to docker hub?

File: Makefile [hunk 2]
```diff
@@ -33,11 +45,14 @@ cleanrun: clean run
...
+
+install-dev: install
+	pip -r requirements-dev.txt
 
-# Setup virtuel environment
+# Setup virtual environment
 venv:
 	python -m venv .pyenv
-	.pyenv\Scripts\activate && make install	
+	source .pyenv/bin/activate && make install
```
provinzio, line 56: Does not work on Windows

provinzio (changes requested):

provinzio:
> Maybe add a make rule to push an image to docker hub?

Sure why not

Makefile
jhoogstraat: Can we support Windows and UNIX systems at the same time? There are multiple activate-scripts even for different shells...

provinzio: This might help with that.
https://stackoverflow.com/questions/4058840/makefile-that-distincts-between-windows-and-unix-like-systems

provinzio:
> 
> 
> Maybe it is smarter if we build a new image with GitHub actions when a new commit is pushed to main?
> Otherwise I'd need to handle credentials for docker hub.

Feel free to add a github action to .github/workflows.

jhoogstraat:
> > Maybe it is smarter if we build a new image with GitHub actions when a new commit is pushed to main?
> > Otherwise I'd need to handle credentials for docker hub.
> 
> Feel free to add a github action to .github/workflows.

The action uses two secrets to log in to Docker Hub:
DOCKERHUB_USERNAME and DOCKERHUB_TOKEN

It then builds and pushes an image with the tag provinzio/cointaxman:latest.

provinzio (approved):

provinzio:
Hey @jhoogstraat,
I am sorry that it took me so long to merge your work. You did a splendid job and I am happy to finally merge your work into main.

PR #74, opened 2021-09-08
Add usdt-retry when binance provides no price for btc

jhoogstraat:
When no price data is provided (price == 0.0 in db) for COIN/BTC this pr uses a fallback to USDT.
`make check-db` will check the database for 0.0-prices and tries to fix them with real data.

In my specific case I could fetch prices for all but one singe entry in the db.

### DIFF
--- a/src/price_data.py
+++ b/src/price_data.py
@@ -120,7 +120,12 @@ class PriceData:
 
         if len(data) == 0:
             log.warning("Binance offers no price for `%s` at %s", symbol, utc_time)
-            return decimal.Decimal()
+            if quote_asset == "USDT":
+                return decimal.Decimal()
+            log.warning(f"Trying {base_asset}USDT and {quote_asset}USDT")
+            usdt = self.get_price("binance", base_asset, utc_time, "USDT")
+            quote = self.get_price("binance", quote_asset, utc_time, "USDT")
+            return usdt / quote
 
         # Calculate average price.
         total_cost = decimal.Decimal()
@@ -563,3 +568,45 @@ class PriceData:
         if isinstance(tr, transaction.SoldCoin):
             return price * tr.sold
         raise NotImplementedError
+
+    def check_database(
+        self,
+    ):
+        for db_path in Path(config.DATA_PATH).glob("*.db"):
+            if db_path.is_file():
+                platform = db_path.stem
+
+                try:
+                    get_price = getattr(self, f"_get_price_{platform}")
+                except AttributeError:
+                    if platform == "coinbase":
+                        get_price = self._get_price_coinbase_pro
+                    else:
+                        raise NotImplementedError(
+                            "Unable to read data from %s", platform
+                        )
+
+                with sqlite3.connect(db_path) as conn:
+                    query = "SELECT name FROM sqlite_master where type='table'"
+                    cur = conn.execute(query)
+
+                    for table in cur.fetchall():
+                        tablename = table[0]
+                        pair = tablename.split("/")
+                        query = f"SELECT utc_time FROM `{tablename}` WHERE price<=0.0;"
+                        cur = conn.execute(query)
+
+                        for row in cur.fetchall():
+                            utc_time = datetime.datetime.strptime(
+                                row[0], "%Y-%m-%d %H:%M:%S%z"
+                            )
+                            price = get_price(pair[0], utc_time, pair[1])
+                            log.warning(
+                                f"Updating {tablename} at {utc_time} to {price}"
+                            )
+                            query = (
+                                f"UPDATE `{tablename}` SET price=? WHERE utc_time=?;"
+                            )
+
+                            conn.execute(query, (str(price), utc_time))
+                            conn.commit()
--- a/src/price_data.py
+++ b/src/price_data.py
@@ -122,9 +122,10 @@ class PriceData:
             log.warning("Binance offers no price for `%s` at %s", symbol, utc_time)
             if quote_asset == "USDT":
                 return decimal.Decimal()
-            log.warning(f"Trying {base_asset}USDT and {quote_asset}USDT")
+            log.info(f"Trying {base_asset}USDT and {quote_asset}USDT")
+            if quote := self.get_price("binance", quote_asset, utc_time, "USDT") == 0.0:
+                return quote
             usdt = self.get_price("binance", base_asset, utc_time, "USDT")
-            quote = self.get_price("binance", quote_asset, utc_time, "USDT")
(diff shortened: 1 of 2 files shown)


scientes:
If im correct this uses usdt as a backup route when btc does not provide a route if I'm correct?  also: https://github.com/provinzio/CoinTaxman/pull/16 should fix and optimize price discovery but uses the current function as a fallback if everything fails

jhoogstraat:
Exactly. 
Until #16 is merged, this can serve as a backup solution.

src/price_data.py
```diff
@@ -120,7 +120,12 @@ def _get_price_binance(
 
         if len(data) == 0:
             log.warning("Binance offers no price for `%s` at %s", symbol, utc_time)
-            return decimal.Decimal()
+            if quote_asset == "USDT":
+                return decimal.Decimal()
+            log.warning(f"Trying {base_asset}USDT and {quote_asset}USDT")
+            usdt = self.get_price("binance", base_asset, utc_time, "USDT")
+            quote = self.get_price("binance", quote_asset, utc_time, "USDT")
```
provinzio, line 127: If `quote_asset : USDT` doesn't exist, this function returns decimal.Decimal(). Which makes usdt/quote results in a `decimal.InvalidOperation: [<class 'decimal.DivisionUndefined'>]`.

We could query `quote` first, and catch the case when `quote_asset : USDT` does not exist.

src/price_data.py
```diff
@@ -563,3 +568,45 @@ def get_cost(
         if isinstance(tr, transaction.SoldCoin):
             return price * tr.sold
         raise NotImplementedError
+
+    def check_database(
+        self,
```
provinzio, line 573: Remove the comma, keeping the function definition a one liner.

src/price_data.py
```diff
@@ -563,3 +568,45 @@ def get_cost(
...
+                    else:
+                        raise NotImplementedError(
+                            "Unable to read data from %s", platform
+                        )
+
+                with sqlite3.connect(db_path) as conn:
+                    query = "SELECT name FROM sqlite_master where type='table'"
+                    cur = conn.execute(query)
+
+                    for table in cur.fetchall():
```
provinzio, line 593: `table` is a confusing variable name. Perhaps call it `result` or something like that instead.

Or use a generator expression `tablenames = (r[0] for r in cur.fetchall()); for tablename in tablenames ...`

src/price_data.py
```diff
@@ -563,3 +568,45 @@ def get_cost(
...
+                        tablename = table[0]
+                        pair = tablename.split("/")
```
provinzio, line 595: We could add a assert here, `assert len(pair) == 2, f"Pair must contain exactly two coins, but {len(pair)} given: {pair}`. Or use `base_asset, quote_asset = tablename.split("/")` instead.

src/price_data.py
```diff
@@ -563,3 +568,45 @@ def get_cost(
...
+                        query = f"SELECT utc_time FROM `{tablename}` WHERE price<=0.0;"
+                        cur = conn.execute(query)
+
+                        for row in cur.fetchall():
+                            utc_time = datetime.datetime.strptime(
+                                row[0], "%Y-%m-%d %H:%M:%S%z"
+                            )
+                            price = get_price(pair[0], utc_time, pair[1])
+                            log.warning(
```
provinzio, line 604: This shouldn't be a warning, merely an info, because this is exactly what the function is good for.

But we could warn at the end, if and how many rows are still `price<=0.0`, perhaps separated by platforms (total: 10 rows, binance: 5 rows, ...)

provinzio (changes requested):
Hey @jhoogstraat, great idea with the `check_database`. Using USDT as additional fallback might be of help as long as #16 isn't merged yet.

src/price_data.py
jhoogstraat: log.info does not show by default it seems.

src/price_data.py
```diff
@@ -587,26 +588,36 @@ def check_database(
                         )
 
                 with sqlite3.connect(db_path) as conn:
-                    query = "SELECT name FROM sqlite_master where type='table'"
+                    query = f"SELECT name FROM sqlite_master where type='table' AND name LIKE '%{config.FIAT_CLASS.name}'"
```
provinzio: Why is `LIKE '%{config.FIAT_CLASS.name}` needed?

provinzio (approved):

src/price_data.py
jhoogstraat: In the end it only matters that all COIN/FIAT_CLASS data is present, no?
So we only need to check if those pairs are ok.

provinzio: We need to have all prices (if existing) independent of the home fiat.

jhoogstraat: Yeah, just noticed myself. Reverting.

src/price_data.py
```diff
@@ -119,8 +119,22 @@ def _get_price_binance(
...
-            log.warning("Binance offers no price for `%s` at %s", symbol, utc_time)
-            return decimal.Decimal()
+            log.warning(
+                f"""
+            Binance offers no price
+            for {symbol} at {utc_time}
+            Retrying with {base_asset}USDT and {quote_asset}USDT
+            """,
+                symbol,
+                utc_time,
```
provinzio, line 129: Use %-formatting in logging. Check other logging messages , too.

src/price_data.py
```diff
@@ -563,3 +577,61 @@ def get_cost(
...
+                                WHERE utc_time=?;"""
+                                conn.execute(query, (str(price), utc_time))
+                                stats[platform]["fix"] += 1
+
+                    conn.commit()
+
+        log.info("Check Database Result:")
+        for platform, result in stats.items():
+            fixed, remaining = result.values()
+            log.info(f"{platform}:\nFixed: {fixed}, Remaining: {remaining}")
```
provinzio, line 637: Unnecessary \n

provinzio (approved):

src/price_data.py
jhoogstraat: Its mixed actually. Line 189 for example. Can you explain the reasoning, so I can infer the correct formatting?

provinzio: %-formatting is lazy evaluated and postpones the string calculation of the argument after the check whether the logging message is necessary (e.g. logging level is warning, a log to debug isn't evaluated).

Most string calculations are very cheap, so it's not that of an performance issue, but I am still hanging on to it, so I never could have the performance problem.

Issue #70, opened 2021-08-31
Bitpanda exchange


sebastianelsner:
I have some old Bitpanda (not Pro) trades. So, I would like to add the exchange. I can add this myself using the code from "pro" as a template. But it would help a lot if someone could post one actual data line of the bitpanda pro csv export for orientation! Much appreciated!

provinzio:
@uvok might be able to help you with that.

uvok:
Sure thing.

For reference, Bitpanda Pro CSV files (as of May 2021, haven't checked this in a while, but the code is based on this).

For transactions (what goes in, what goes out of the account):

```
"Disclaimer: All data is without guarantee, errors and changes are reserved."
"Account id : aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"

"Bitpanda Pro transactions history"
"Transaction ID","External Transaction ID",Type,In/Out,Internal/External,Amount,Currency,Fee,"Time Created"
aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee,n/a,Deposit,Incoming,External,100.00000000,EUR,0.00000000,2021-01-01T00:00:00+02:00
aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeef,n/a,Withdrawal,Outgoing,External,200.00000000,EUR,0.00000000,2021-01-01T00:00:00+02:00
```

For trades:

```
"Disclaimer: All data is without guarantee, errors and changes are reserved."
Account ID: aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee
Bitpanda Pro trade history
Order ID,Trade ID,Type,Market,Amount,Amount Currency,Price,Price Currency,Fee,Fee Currency,Time (UTC)
aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeed,aaaaaaab-bbbb-cccc-dddd-eeeeeeeeeeee,SELL,BTC_EUR,0.00100,BTC,50000,EUR,0.01000000,BEST,2021-05-01T00:00:00.000000Z
aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeec,aaaaaaab-bbbb-cccc-dddd-eeeeeeeeeeef,BUY,DOT_EUR,3.0000,DOT,30.000,EUR,0.05555555,BEST,2021-05-01T08:00:00.000000Z
```

uvok:
*sighs*

[...]

PR #76, opened 2021-09-12
add simple Bitpanda exchange

sebastianelsner:
Hi, I have added a csv import for the simple bitpanda exchange, taking bitpanda pro as inspiration,  please review. Thanks!

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -678,7 +678,7 @@ class Book:
 
             for (
                 _tx_id,
-                utc_time,
+                csv_utc_time,
                 operation,
                 _inout,
                 amount_fiat,
@@ -697,23 +697,27 @@ class Book:
                 row = reader.line_num
 
                 # make RFC3339 timestamp ISO 8601 parseable
-                if utc_time[-1] == "Z":
-                    utc_time = utc_time[:-1] + "+00:00"
+                if csv_utc_time[-1] == "Z":
+                    csv_utc_time = csv_utc_time[:-1] + "+00:00"
 
                 # timezone information is already taken care of with this
-                utc_time = datetime.datetime.fromisoformat(utc_time)
+                utc_time = datetime.datetime.fromisoformat(csv_utc_time)
 
                 # transfer ops seem to be akin to airdrops. In my case I got a
                 # CocaCola transfer, which I don't want to track. Would need to
                 # be implemented if need be.
-                if operation in ["transfer"]:
+                if operation == "transfer":
                     log.warning(
                         f"'Transfer' operations are not "
                         f"implemented, skipping: {file_path} line {row}."
                     )
                     continue
 
-                operation = operation_mapping.get(operation)
+                # fail for unknown ops
+                try:
+                    operation = operation_mapping[operation]
+                except KeyError:
+                    raise RuntimeError(f"Unsupported operation '{operation}'")
 
                 if operation == "Deposit":
                     # This identifies a fiat deposit. Error on yet unknown deposits,
@@ -740,8 +744,6 @@ class Book:
                     self.price_data.set_price_db(
                         platform, asset, "EUR", utc_time, price
                     )
-                else:
-                    raise RuntimeError(f"Unsupported operation '{operation}'")
 
                 if change < 0:
                     raise RuntimeError(
--- a/src/book.py
+++ b/src/book.py
@@ -627,11 +627,158 @@ class Book:
                     file_path,
                 )
 
+    def _read_bitpanda(self, file_path: Path) -> None:
+        """Reads a trade statement from Bitpanda.
+
+        Args:
+            file_path (Path): Path to Bitpanda trade history.
+        """
+
+        platform = "bitpanda"
+
+        operation_mapping = {
+            "deposit": "Deposit",
+            "withdrawal": "Withdraw",
+            "buy": "Buy",
+            "sell": "Sell",
+        }
+
+        with open(file_path, encoding="utf8") as f:
+            reader = csv.reader(f)
+
+            # skip header, there are multiple lines
+            # the last one is the actual header
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+
+            if line != [
+                "Transaction ID",
+                "Timestamp",
+                "Transaction Type",
+                "In/Out",
+                "Amount Fiat",
+                "Fiat",
+                "Amount Asset",
+                "Asset",
+                "Asset market price",
+                "Asset market price currency",
+                "Asset class",
+                "Product ID",
+                "Fee",
+                "Fee asset",
+                "Spread",
+                "Spread Currency",
+            ]:
+                raise RuntimeError(f"Expected header did not match {line}")
+
+            for (
+                _tx_id,
+                utc_time,
+                operation,
+                _inout,
+                amount_fiat,
+                fiat,
+                amount_asset,
+                asset,
+                asset_price,
+                asset_price_currency,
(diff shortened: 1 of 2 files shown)


src/book.py
```diff
@@ -627,11 +627,160 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                # fail for unknown ops
+                try:
+                    operation = operation_mapping[operation]
+                except KeyError:
+                    raise RuntimeError(f"Unsupported operation '{operation}'")
+
+                if operation == "Deposit":
+                    # This identifies a fiat deposit. Error on yet unknown deposits,
+                    # they will need to be implemented separately
+                    if fiat != "EUR" and asset != "EUR" and asset_class != "fiat":
```
provinzio, line 725: Why is this important to distinguish? Deposits doesnt matter for taxation stuff. They are only there for information.

src/book.py
```diff
@@ -627,11 +627,160 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                        raise RuntimeError(
+                            "A deposit is expected to be of fiat and 'EUR' currency. "
+                            "Others are not implemented yet."
+                        )
+                    else:
+                        change = misc.force_decimal(amount_fiat)
+                elif operation == "Withdraw":
+                    change = misc.force_decimal(amount_asset)
+                elif operation in ["Buy", "Sell"]:
+                    if asset_price_currency != "EUR":
```
provinzio, line 735: Check against home fiat constant.

src/book.py
```diff
@@ -627,11 +627,160 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                        raise RuntimeError(
+                            "Only Euro is supported as 'Asset market price currency' "
+                            "currency, since price fetching is not fully implemented "
```
provinzio, line 738: "price fetching for fiat currencies"

src/book.py
```diff
@@ -627,11 +627,160 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                    change = misc.force_decimal(amount_asset)
+                    # Save price in our local database for later.
+                    price = misc.force_decimal(asset_price)
+                    self.price_data.set_price_db(
+                        platform, asset, "EUR", utc_time, price
+                    )
+
+                if change < 0:
+                    raise RuntimeError(
+                        f"Unexpected value for 'Amount Asset' column {change}"
```
provinzio, line 750: change could also be from amount fiat (line 731)

src/book.py
```diff
@@ -627,11 +627,160 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                        row,
+                        file_path,
+                    )
+
     def detect_exchange(self, file_path: Path) -> Optional[str]:
         if file_path.suffix == ".csv":
             with open(file_path, encoding="utf8") as f:
                 reader = csv.reader(f)
-                header = next(reader, None)
+                csv_content = list(reader)
```
provinzio, line 772: Can we accomplish this without reading the whole csv file into memory?

src/book.py
sebastianelsner: I see, my reasoning is that if I don't know the full details of the deposit as a spec, it can be "anything", other currencies, "airdrops" (?) and thus it should fail hard immediately as it's unexpected. But if you say its unreasonable I am open to just remove it.

src/book.py
sebastianelsner: you mean config.FIAT ?

src/book.py
sebastianelsner: sure

src/book.py
provinzio: Jep

src/book.py
```diff
@@ -627,11 +627,158 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+
+            # skip header, there are multiple lines
+            # the last one is the actual header
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
+            line = next(reader)
```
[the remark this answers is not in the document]
sebastianelsner: I just did a re-export and i get:


Disclaimer: All data is without guarantee, errors and changes are reserved.
My Name, 1970-01-1
myname@mail.ru
Account opened at: 4/9/88, 1:33 PM
Venue: Bitpanda
Reported by Bitpanda GmbH
Transaction ID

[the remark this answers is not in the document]
Griffsano: Except for this, the code works fine with my Bitpanda export. Thanks for implementing!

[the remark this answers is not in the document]
sebastianelsner: So you dont get the "Reported by Bitpanda GmbH" Line? Did you do the export via "Verlauf"?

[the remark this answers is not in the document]
Griffsano: > I just did a re-export and i get:
> 
> Disclaimer: All data is without guarantee, errors and changes are reserved. My Name, 1970-01-1 [myname@mail.ru](mailto:myname@mail.ru) Account opened at: 4/9/88, 1:33 PM Venue: Bitpanda Reported by Bitpanda GmbH Transaction ID

That's odd, for me the "Reported by Bitpanda GmbH" line is missing.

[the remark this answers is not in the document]
Griffsano: > So you dont get the "Reported by Bitpanda GmbH" Line? Did you do the export via "Verlauf"?

Yep. Top right corner -> History/Verlauf -> Export/Exportieren -> Generate CSV / CSV generieren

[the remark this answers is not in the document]
Griffsano: I have no idea why we're getting two different CSV versions. What do you think of:
```suggestion
            if line == "Reported by Bitpanda GmbH":
                line = next(reader)
```
This should work for both versions.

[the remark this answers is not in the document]
provinzio: Wouldn't it be possible to skip lines until the first line of the table start?

src/book.py
provinzio: It is also possible to have coins deposited (transfered from somewhere else aka withdrawn from somewhere and deposited here). This is also not taxable. 

It would be a huge problem if they declare airdrops as deposits and I really hope that they don't do that...
If they declare airdrops as deposit

src/book.py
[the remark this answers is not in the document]
Griffsano: What do you think about this?:
```
...
        with open(file_path, encoding="utf8") as f:
            reader = csv.reader(f)
            line = next(reader)

            # skip header, there are multiple lines
            while line != [
                "Transaction ID",
                "Timestamp",
                "Transaction Type",
                "In/Out",
                "Amount Fiat",
                "Fiat",
                "Amount Asset",
                "Asset",
                "Asset market price",
                "Asset market price currency",
                "Asset class",
                "Product ID",
                "Fee",
                "Fee asset",
                "Spread",
                "Spread Currency",
            ]:
                try:
                    line = next(reader)
                except StopIteration:
                    raise RuntimeError(f"Expected header not found in file {file_path}")

            for (
...
```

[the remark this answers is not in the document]
sebastianelsner: yep, lets do something like this. will have time on the weekend.

src/book.py
Griffsano: I think we can remove this check completely, as it shouldn't matter which fiat currency it is. I just tried replacing EUR with USD in my CSV (USDT for fee asset), and everything still worked as intended. My proposed changes are here: https://github.com/sebastianelsner/CoinTaxman/pull/1

Griffsano:
To link the issue: Closes #70

provinzio:
This PR is stale and worked on in #101.

PR #78, opened 2021-09-14
add new binance format

### LINKED ISSUE #77: Unable to detect the exchange of file
I get an "Unable to detect the exchange of file" when trying to start the program?

My account_statement is from binance. Is there anything else apart from moving the CSV file to the account_statements folder that I need to do before starting?

scientes:
should fix #77 but haven't tested it yet fully

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -66,7 +66,7 @@ class Book:
         o = Op(utc_time, platform, change, coin, row, file_path)
         self.operations.append(o)
 
-    def _read_binance(self, file_path: Path, version=1) -> None:
+    def _read_binance(self, file_path: Path, version: int = 1) -> None:
         platform = "binance"
         operation_mapping = {
             "Distribution": "Airdrop",
@@ -88,7 +88,7 @@ class Book:
 
             for rowlist in reader:
                 if version == 1:
-                    _utc_time, account, operation, coin, _change, remark = reader
+                    _utc_time, account, operation, coin, _change, remark = rowlist
                 if version == 2:
                     (
                         user_ID,
@@ -98,9 +98,9 @@ class Book:
                         coin,
                         _change,
                         remark,
-                    ) = reader
+                    ) = rowlist
                 else:
-                    log.warning("File version not Supported " + file_path)
+                    log.warning("File version not Supported " + str(file_path))
                     break
 
                 row = reader.line_num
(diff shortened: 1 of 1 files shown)


src/book.py
```diff
@@ -86,7 +86,23 @@ def _read_binance(self, file_path: Path) -> None:
             # Skip header.
             next(reader)
 
-            for _utc_time, account, operation, coin, _change, remark in reader:
+            for rowlist in reader:
+                if version == 1:
+                    _utc_time, account, operation, coin, _change, remark = rowlist
+                if version == 2:
+                    (
+                        user_ID,
```
provinzio, line 94: I'd prefer _ instead of user Id

src/book.py
```diff
@@ -86,7 +86,23 @@ def _read_binance(self, file_path: Path) -> None:
...
+                        _utc_time,
+                        account,
+                        operation,
+                        coin,
+                        _change,
+                        remark,
+                    ) = rowlist
+                else:
+                    log.warning("File version not Supported " + str(file_path))
```
provinzio, line 103: This will spam your Log messages. Each line will write a warning.
Just raise Not implemented error

src/book.py
```diff
@@ -125,6 +141,9 @@ def _read_binance(self, file_path: Path) -> None:
                     operation, utc_time, platform, change, coin, row, file_path
                 )
 
+    def _read_binance_new(self, file_path: Path) -> None:
```
provinzio, line 144: Don't name it new. Someday it will change again and we won't name the next one newer :D

Maybe read binance 2. And it's version 2?

provinzio (changes requested):

src/book.py
scientes: this should only raise one per file due to the break afterwords

src/book.py
scientes: i dont know if binance has a versioning scheme

provinzio:
Looks good to me. What do you think?

src/book.py
```diff
@@ -86,7 +86,23 @@ def _read_binance(self, file_path: Path) -> None:
             # Skip header.
             next(reader)
 
-            for _utc_time, account, operation, coin, _change, remark in reader:
+            for rowlist in reader:
+                if version == 1:
+                    _utc_time, account, operation, coin, _change, remark = reader
+                if version == 2:
```
Griffsano, line 92: This should read `elif version == 2:`. Otherwise, if the file is version 1, it would trigger the warning/error from the else block below.

provinzio:
Finished in #88

Issue #50, opened 2021-06-15
[Binance] Calculate price from transaction statement


scientes:
Currently the price for a certain transaction statement is currently not known for binance and some other exchanges.
but for binance transactions that belong together always have the correct timestamp if i'm not mistaken.
```
2021-05-09 03:25:21,Spot,Transaction Related,USDT,-21.87041400,""
2021-05-09 03:25:21,Spot,Fee,BNB,-0.00002523,""
2021-05-09 03:25:21,Spot,Buy,VET,98.10000000,""
```
Thus the absolute of Transaction Related divided by buy should give us the price of VET/USDT without requiring an api request.
98.1/21.8741400=0.2294

This does not work for all statements but should improve the accuracy of some and the speed (if we have eur as base fiat we'd need to compute the USDT/EUR but we would save probably 1 api request)

provinzio:
done with the above PR.

PR #96, opened 2021-12-13
calculate prices from csv

scientes:
This pr does 4 Things:

1. updates the year to 2021
2. All values are now stored as str in db to prevent floating point bugs in the future (there is no decimal in sqlite)
3. Tablenames are now Alphabeticaly sorted ETH/BNB -> BNB/ETH and converted to prevent two tables being created for the same pair
4. If there are two timestamps from the same platform the price between them is calculated when one is a buy operation and the other is a sell operation (#50)

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -696,6 +696,37 @@ class Book:
 
         return None
 
+    def get_price_from_csv(self):
+        print("hi")
+        for platform, operations_a in misc.group_by(
+            self.operations, "platform"
+        ).items():
+            for timestamp, operations_b in misc.group_by(
+                operations_a, "utc_time"
+            ).items():
+                if len(operations_b) > 1:
+                    buytransaction = selltransaction = None
+                    for operation in operations_b:
+                        if isinstance(operation, tr.Buy):
+                            buytransaction = operation
+                        elif isinstance(operation, tr.Sell):
+                            selltransaction = operation
+                        if buytransaction is not None and selltransaction is not None:
+                            price = decimal.Decimal(
+                                selltransaction.change / buytransaction.change
+                            )
+                            logging.debug(
+                                f"Added price from csv: {selltransaction.coin}/{buytransaction.coin} price: {price}"
+                            )
+                            self.price_data.set_price_db(
+                                platform,
+                                selltransaction.coin,
+                                buytransaction.coin,
+                                timestamp,
+                                price,
+                            )
+                            break
+
     def read_file(self, file_path: Path) -> None:
         """Import transactions form an account statement.
 
--- a/src/book.py
+++ b/src/book.py
@@ -27,6 +27,7 @@ import misc
 import transaction as tr
 from core import kraken_asset_map
 from price_data import PriceData
+from database import set_price_db
 
 log = logging.getLogger(__name__)
 
@@ -284,7 +285,7 @@ class Book:
                 assert _currency_spot == "EUR"
 
                 # Save price in our local database for later.
-                self.price_data.set_price_db(platform, coin, "EUR", utc_time, eur_spot)
+                set_price_db(platform, coin, "EUR", utc_time, eur_spot)
 
                 if operation == "Convert":
                     # Parse change + coin from remark, which is
@@ -317,7 +318,7 @@ class Book:
                     )
 
                     # Save convert price in local database, too.
-                    self.price_data.set_price_db(
+                    set_price_db(
                         platform, convert_coin, "EUR", utc_time, convert_eur_spot
                     )
                 else:
@@ -721,9 +722,9 @@ class Book:
 
                 # Save price in our local database for later.
                 price = misc.force_decimal(_price)
-                self.price_data.set_price_db(platform, coin, "EUR", utc_time, price)
+                set_price_db(platform, coin, "EUR", utc_time, price)
                 if best_price:
-                    self.price_data.set_price_db(
+                    set_price_db(
                         platform,
                         "BEST",
                         "EUR",
@@ -885,7 +886,7 @@ class Book:
                     f"{price} for {platform} at {timestamp}"
                 )
 
-                self.price_data.set_price_db(
+                set_price_db(
                     platform,
                     buytr.coin,
                     selltr.coin,
--- a/src/book.py
+++ b/src/book.py
@@ -853,9 +853,11 @@ class Book:
                             sellcount += 1
 
                     if buycount == 1 and sellcount == 1:
+                        # price definition example: BTCEUR = traded EUR / traded BTC
                         price = decimal.Decimal(buytr.change / selltr.change)
                         logging.debug(
-                            f"Added price from csv: {selltr.coin}/{buytr.coin} price: {price} timestamp: {timestamp}"
+                            f"Added {selltr.coin}/{buytr.coin} price from CSV: "
+                            f"{price} for {platform} at {timestamp}"
                         )
                         self.price_data.set_price_db(
                             platform,
--- a/src/book.py
+++ b/src/book.py
@@ -905,6 +905,8 @@ class Book:
         assert file_path.is_file()
 
         if exchange := self.detect_exchange(file_path):
+            # TODO check that database file exists. if missing, add file with
+            #      highest version number (highest patch number)
             try:
                 read_file = getattr(self, f"_read_{exchange}")
             except AttributeError:
--- a/src/book.py
+++ b/src/book.py
@@ -834,38 +834,64 @@ class Book:
 
         return None
 
-    def get_price_from_csv(self):
-        for platform, operations_a in misc.group_by(
+    def get_price_from_csv(self) -> None:
+        """Calculate coin prices from buy/sell operations in CSV files.
+
+        When exactly one buy and sell happend at the exact same time,
+        these two operations might belong together and we can calculate
+        the paid price for this transaction.
+        """
+        # Group operations by platform.
+        for platform, platform_operations in misc.group_by(
             self.operations, "platform"
         ).items():
-            for timestamp, operations_b in misc.group_by(
-                operations_a, "utc_time"
+            # Group operations by time.
+            # Look at all operations which happend at the same time.
+            for timestamp, time_operations in misc.group_by(
+                platform_operations, "utc_time"
             ).items():
-                if len(operations_b) > 1:
-                    buytr = selltr = None
-                    buycount = sellcount = 0
-                    for operation in operations_b:
-                        if isinstance(operation, tr.Buy):
-                            buytr = operation
-                            buycount += 1
-                        elif isinstance(operation, tr.Sell):
-                            selltr = operation
-                            sellcount += 1
+                buytr = selltr = None
+                buycount = sellcount = 0
 
-                    if buycount == 1 and sellcount == 1:
-                        # price definition example: BTCEUR = traded EUR / traded BTC
-                        price = decimal.Decimal(buytr.change / selltr.change)
-                        logging.debug(
-                            f"Added {selltr.coin}/{buytr.coin} price from CSV: "
-                            f"{price} for {platform} at {timestamp}"
-                        )
-                        self.price_data.set_price_db(
-                            platform,
-                            selltr.coin,
-                            buytr.coin,
-                            timestamp,
-                            price,
-                        )
+                # Extract the buy and sell operation.
+                for operation in time_operations:
+                    if isinstance(operation, tr.Buy):
+                        buytr = operation
+                        buycount += 1
+                    elif isinstance(operation, tr.Sell):
+                        selltr = operation
+                        sellcount += 1
+
+                # Skip the operations of this timestamp when there aren't
+                # exactly one buy and one sell operation.
+                # We can only match the buy and sell operations, when there
+                # are exactly one buy and one sell operation.
+                if not (buycount == 1 and sellcount == 1):
+                    continue
+
+                assert isinstance(timestamp, datetime.datetime)
+                assert isinstance(buytr, tr.Buy)
+                assert isinstance(selltr, tr.Sell)
+
+                # Price definition example for buying BTC with EUR:
+                # Symbol: BTCEUR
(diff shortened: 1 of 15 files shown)


src/price_data.py
```diff
@@ -646,7 +659,10 @@ def get_price(
             # Do not save price in database.
             price = self.__mean_price_db(db_path, tablename, utc_time)
 
-        return price
+        if inverted:
+            return decimal.Decimal(1 / price)
```
provinzio, line 663: You can use `misc.reciprocal` for that.
https://github.com/provinzio/CoinTaxman/blob/main/src/misc.py#L90-L91

src/price_data.py
```diff
@@ -627,15 +633,22 @@ def get_price(
...
-        tablename = self.get_tablename(coin, reference_coin)
+        if coin < reference_coin:
+            coin_a = coin
+            coin_b = reference_coin
+            inverted = False
+        else:
+            coin_a = reference_coin
+            coin_b = coin
+            inverted = True
+        tablename = self.get_tablename(coin_a, coin_b)
```
provinzio, line 644: As this logic is also used in set_price, we should create a function which returns all these... something like

```
def get_XYZ(coin, reference_coin) -> (coin_a, coin_b, tablename, reciprocal):
  if reciprocal := coin > reference_coin:
    ...
  else:
    ....
  tablename = self.get_tablename(coin, reference_coin)

```

provinzio (changes requested):
Hey @scientes,

good idea to group the tables of opposite coin pairs. I am currently unsure, if this is a good idea. e.g. BTC/USD is like 60.000$ but USD/BTC would be 0.00001666666 which is unnecessary small.

Is it enough to save the decimals as string for that? Please check this explicitly.

Griffsano:
Hey, I think the inversion of prices will get problematic for coins that are listed for 0 EUR (e.g. airdrops of some new tokens). Would it be possible to search the database for the inverse coin pair if the original cannot be found, before creating a new table?

provinzio:
> Hey, I think the inversion of prices will get problematic for coins that are listed for 0 EUR (e.g. airdrops of some new tokens). Would it be possible to search the database for the inverse coin pair if the original cannot be found, before creating a new table?

If the coin pair ABC/XYZ is valued at 0 Eur, the coin pair XYZ/ABC should be 0 Eur as well. Therefore creating the reciprocal requies to except the case when the pair is valued at 0 Eur. The reciprocal function in misc.py does exactly that. I mentioned it in the code review. 

Furthermore, this PR sorts the coin pair alphabetically to form the table name. Therefore, each coin pair plus his "inverse coin pair" will have exactly one table. There wouldn't be the need to search for another table, as there is only this one.

src/price_data.py
```diff
@@ -550,7 +550,7 @@ def __set_price_db(
                     create_query = (
                         f"CREATE TABLE `{tablename}`"
                         "(utc_time DATETIME PRIMARY KEY, "
-                        "price FLOAT NOT NULL);"
+                        "price STR NOT NULL);"
```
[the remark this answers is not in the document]
provinzio: The patch function should also reorganize our tables, so that all coin pairs are sorted alphabetically. This might also require to merge tables.

Griffsano:
You're right, it should also work with this function, basically overwriting inf by 0. Of course, if the inverse price can always be computed, the alphabetic sorting works and we don't have to search for the inverse coin pair in the database.

scientes:
> Hey @scientes,
> 
> good idea to group the tables of opposite coin pairs. I am currently unsure, if this is a good idea. e.g. BTC/USD is like 60.000$ but USD/BTC would be 0.00001666666 which is unnecessary small.
> 
> Is it enough to save the decimals as string for that? Please check this explicitly.

The default precision of decimal is 28
also small test:
```
>>> Decimal(1)/(Decimal(1) /Decimal(123456789123456789))
Decimal('123456789123456789.0000000000')
>>> Decimal(1)/(Decimal(1) /Decimal(123456789123456789.1))
Decimal('123456789123456784.0000000000')
>>> Decimal(1)/(Decimal(1) /Decimal(12345.1))
Decimal('12345.10000000000036379788071')
```
the max of string length in sqllite is 2^31-1 bytes which i think is more than enough

I'm not super sure how to fix that properly, i mean it's still more accurate than fetching the 1 minute average. The easiest would be just to up the precision on decimal. or we could make two columns in the db one for the price and one for if its inverted relative to the table name and we do not invert the price at all

src/book.py
```diff
@@ -834,6 +834,35 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
 
         return None
 
+    def get_price_from_csv(self):
+        print("hi")
```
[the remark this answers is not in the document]
scientes: 'Caveman Debugging' that i haven't removed :D

src/book.py
```diff
@@ -834,6 +834,35 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
+            for timestamp, operations_b in misc.group_by(
+                operations_a, "utc_time"
+            ).items():
+                if len(operations_b) > 1:
+                    buytr = selltr = None
+                    for operation in operations_b:
+                        if isinstance(operation, tr.Buy):
+                            buytr = operation
+                        elif isinstance(operation, tr.Sell):
+                            selltr = operation
```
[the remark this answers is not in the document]
scientes: ah forgot that case. In case of multiple buy and sell i would just skip that pair, because there is not enough information on which buy and sell belongs to which pair.
```csv
2021-06-09 06:32:58,Spot,Small assets exchange BNB,BNB,9.9E-7,""
2021-06-09 06:32:58,Spot,Small assets exchange BNB,BNB,0.00000347,""
2021-06-09 06:32:58,Spot,Small assets exchange BNB,ATA,-0.00547803,""
2021-06-09 06:32:58,Spot,Small assets exchange BNB,FRONT,-0.00040334,""
2021-06-09 06:32:58,Spot,Small assets exchange BNB,BNB,0.00001395,""
2021-06-09 06:32:58,Spot,Small assets exchange BNB,LTC,-0.00000784,""
2021-06-09 13:36:59,Spot,Small assets exchange BNB,BNB,9.9E-7,""
2021-06-09 13:36:59,Spot,Small assets exchange BNB,FRONT,-0.00040621,""
```

provinzio:
The precision should be high enough as it is right now. At least I don't know of a coin pair with such a huge factor. Binance shows the prices with 8 decimals places (at least in the GUI).

scientes:
@provinzio I added the migrations to the check_db function. but apparently its not used anywhere. is that the correct place for the migration or should i make an extra function for them?

src/book.py
```diff
@@ -834,6 +834,37 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
+                    for operation in operations_b:
+                        if isinstance(operation, tr.Buy):
+                            buytr = operation
+                            buycount += 1
+                        elif isinstance(operation, tr.Sell):
+                            selltr = operation
+                            sellcount += 1
+
+                    if buycount == 1 and sellcount == 1:
+                        price = decimal.Decimal(buytr.change / selltr.change)
```
Griffsano, line 856: Shouldn't this be sell/buy?

scientes: Yes but it works and i dunno why if we use sell/buy all prices are inverted in the DB (eur/btc=30k) and if we use buy/sell there Not inverted (eur/btc=0.00XXX) and i havent found out why yet

Griffsano: I'm also confused now.
Could it be that the prices are actually inverted in the database? If one BTC costs 30k EUR, then BTC/EUR should be 1/30k, right? But the database stores it the other way around.

Griffsano: If I use `price = decimal.Decimal(selltr.change / buytr.change)` and replace `if reciprocal` by `if not reciprocal` in `price_data.py` (line 661), then I get correct calculations for both the CSV prices and the API prices (tested with Kraken).

scientes: > I'm also confused now. Could it be that the prices are actually inverted in the database? If one BTC costs 30k EUR, then BTC/EUR should be 1/30k, right? But the database stores it the other way around.

Btc/eur should be 30k if btc costs 30k

Griffsano: I see two possible ways of defining the price:
1. The "physical" price definition: Trade 30k EUR for 1 BTC:
BTC/EUR = 1 (BTC) / 30k (EUR) = 1/30k
In this case, the price should be `price = decimal.Decimal(selltr.change / buytr.change)`.
2. The price definition used for charts on various exchanges:
BTC/EUR = price of one BTC in EUR = 30k
In this case `buytr.change / selltr.change` is already correct. But this somehow messes with data that is fetched via the APIs.

src/price_data.py
```diff
@@ -627,15 +639,15 @@ def get_price(
...
         # Check if price exists already in our database.
         if (price := self.__get_price_db(db_path, tablename, utc_time)) is None:
             try:
                 get_price = getattr(self, f"_get_price_{platform}")
             except AttributeError:
                 raise NotImplementedError("Unable to read data from %s", platform)
-
             price = get_price(coin, utc_time, reference_coin, **kwargs)
             assert isinstance(price, decimal.Decimal)
             self.__set_price_db(db_path, tablename, utc_time, price)
```
Griffsano, line 653: Is this missing here?
```
if reciprocal:
    price = misc.reciprocal(price)
```
Since coin/reference_coin are used for fetching the price from the API, the price still needs to be inverted if `coin_a/b` are flipped.

With this change, it should also work if we keep `buytr.change / selltr.change` (for my comment above for `book.py`). I think this is exactly why there is an inconsistency between prices fetched from CSV and from the APIs.

scientes: Looks like it would.

src/book.py
scientes: I'd Stick with 2. Because thats what we get from All APIs.

Griffsano: Alright, then this calculation should be right.

src/price_data.py
```diff
@@ -596,6 +599,15 @@ def set_price_db(
             else:
                 raise e
 
+    def _sort_pair(self, coin: str, reference_coin: str) -> list[Union[str, str, bool]]:
```
scientes, line 602:
```bash
src/price_data.py:585:40: error: Argument 1 to "get_tablename" of "PriceData" has incompatible type "Union[str, str, bool]"; expected "str"  [arg-type]
src/price_data.py:585:48: error: Argument 2 to "get_tablename" of "PriceData" has incompatible type "Union[str, str, bool]"; expected "str"  [arg-type]
src/price_data.py:592:53: error: Argument 2 to "get_price" of "PriceData" has incompatible type "Union[str, str, bool]"; expected "str"  [arg-type]
src/price_data.py:592:71: error: Argument 4 to "get_price" of "PriceData" has incompatible type "Union[str, str, bool]"; expected "str"  [arg-type]
src/price_data.py:643:40: error: Argument 1 to "get_tablename" of "PriceData" has incompatible type "Union[str, str, bool]"; expected "str"  [arg-type]
src/price_data.py:643:48: error: Argument 2 to "get_tablename" of "PriceData" has incompatible type "Union[str, str, bool]"; expected "str"  [arg-type]
```
I'm not sure what the ideal typing would be for this return list/tuple

provinzio:
I added some docstrings and comments and worked on an first idea of a patch system. Unfortunatly, I wasn't able to finish my work and am still unsure whether this is the best approach. I think that we are currently missing a central database class which handles the interaction between the code and our database system. We might want to factor out the database functions from `price_data.py` and some functions from `patch_database.py` and move them to a new `database.py` file. What do you think of that?

The patches in `patch_database.py` aren't finished, but my ground work should give an idea of what is missing. We need a create patches, so that (my) old databases with unsorted tablenames and float prices are converted to the new format with sorted tablenames and string prices.

Can you work on it?

scientes:
yea sure

scientes:
for the db migrations Alembic would be ideal as a 3rd party module but sadly that would require us to switch to sqlalchemy just because we need a db migration function. And thats then kinda overkill

provinzio:
Alembic is a really nice module, but I believe that the benefit of using alembic for our small db is insignificant compared to the effort of getting used with alembic.

scientes:
> patch_database.py

do we wamt to do one class per db file and one in total which contains all db files as subobjects

scientes:
thats all functions that i would move to database.py

provinzio:
> do we wamt to do one class per db file and one in total which contains all db files as subobjects

Mhh, interessant idea. I never thought of implementing it as a database handle, but just factoring out code into a "database utility file". I believe, the easiest way is to have just a database utility file (each function handles the database opening/closing by itself) [1]. The awesome, but complex implementation, would be, to have a database handle class (one should be enough, as all database files work the same), which can be used in a with statement to open/close the database and has the get_price, set_price, ... functions [2]. I don't see why we need a database handle for each class.

Example for of [1]: looks like now, but the functions are now in database_utility.py and not in price_data.py
```
from database_utility import get_price, set_price

platform = 'binance'
pair = 'BTCEUR'
time = now()

price = get_price(platform, pair, time)
price = price + 1
set_price(platform, pair, time)
```

Example use of [2]: one class does the database handling/management
```
platform = 'binance'
pair = 'BTCEUR'
time = now()

with MyDatabaseClass(platform) as db:
    price = db.get_price(pair, time)
    price = price + 1
    db.set_price(price, time)
```
[2] could also be used like so
```
db = MyDatabaseClass(platform)
...
db.close()
```

scientes:
the patch functions somehow don't work because they are not found by their func_prefix

provinzio:
This PR definitly got quite out of scope quickly, but you two, @Griffsano and @scientes, handled it really good, so that we can now calculate the prices directly from CSV files and have an database versioning system!

While reviewing, I noticed, that we got the database type of the price column wrong. "STR" is not supported by sqllite. Please check your databases for the correct type, which should be at least VARCHAR(255). Otherwhise your prices will be stored with a maximal precision of 1e-9.

I added the functionality, that prices from CSV overwrite already existing prices in the database.

Again, good job you two!

Issue #97, opened 2021-12-13
Kraken: Action Types in CSV Exports


Griffsano:
# Summary

The current implementation for reading the Kraken CSV exports is according to the documentation of the "type" field:
https://support.kraken.com/hc/en-us/articles/360001169383-Explanation-of-Ledger-Fields

However, it seems like the CSV exports have changed and do not correspond to the definition above anymore. This leads to completely different behavior for some action types.

For the exports that I have, I can see the following behavior:

# Actions that currently do not work

## Staking / Unstaking

A combination of "withdrawal"/"deposit" and two "transfer" rows.
The current implementation considers "transfer" as airdrop. Depending on how the CSV exports looked before, it could be more or less complicated to guarantee backwards-compatibility. Also related to https://github.com/provinzio/CoinTaxman/issues/57.
For https://github.com/provinzio/CoinTaxman/issues/87, a similar behavior was observed, although slightly different (four "transfer" rows).
For ETH, replace asset with XETH and asset.S with ETH2.S.

### Staking

| txid   | refid   | time      | type         | subtype           | aclass     | asset   | amount   | fee | balance |
| ------ | ------- | --------- | ------------ | ----------------- | ---------- | ------- | -------- | --- | ------- |
| ""     | refid 1 | date time | "withdrawal" | ""                | "currency" | asset   | - amount | fee | ""      |
| ""     | refid 2 | date time | "deposit"    | ""                | "currency" | asset.S | amount   | fee | ""      |
| txid 1 | refid 1 | date time | "transfer"   | "spottostaking"   | "currency" | asset   | - amount | fee | balance |
| txid 2 | refid 2 | date time | "transfer"   | "stakingfromspot" | "currency" | asset.S | amount   | fee | balance |

### Unstaking

| txid   | refid   | time      | type         | subtype           | aclass     | asset   | amount   | fee | balance |
| ------ | ------- | --------- | ------------ | ----------------- | ---------- | ------- | -------- | --- | ------- |
| ""     | refid 1 | date time | "withdrawal" | ""                | "currency" | asset.S | - amount | fee | ""      |
| ""     | refid 2 | date time | "deposit"    | ""                | "currency" | asset   | amount   | fee | ""      |
| txid 1 | refid 2 | date time | "transfer"   | "spotfromstaking" | "currency" | asset   | amount   | fee | balance |
[...]

Issue #57, opened 2021-06-23
Regonize speculation period (Spekulationsfrist) for lending and staking.


provinzio:
According to a draft of the Federal Ministry of Finance of Germany ("Bundesfinanzministerium"), coin lending and staking raises the speculation period ("Spekulationsfrist") from one to ten years.

We shoud adjust the way, staking and lending profits are handled by CoinTaxman.

For me, it stays unclear whether FIFO has to be followed or not.
Example:
1. I buy a coin
2. I buy a second coin
3. I stake one coin: Which coin do I stake? Do I have to use FIFO? Let's assume I stake the first coin.
4. I wait so that both coins are older than 1 year. I am still staking the first coin.
4. I sell a coin: Which coin do I sell? According to FIFO I have to sell the first coin which is currently staked and pay taxes on it because of the increased speculation period. But I'd rather sell the seconds coin, which is actually the only one available for me.

German Source:
https://www.heise.de/news/Fuer-Proof-of-Stake-genutztes-Kryptogeld-erst-nach-10-Jahren-steuerfrei-6114198.html
Draft of BMF: https://www.bundesfinanzministerium.de/Content/DE/Downloads/BMF_Schreiben/Steuerarten/Einkommensteuer/2021-06-17-est-kryptowaehrungen.pdf?__blob=publicationFile&v=2

scientes:
NExt question is what happens with defi stuff like swaps, does that apply there too?

Griffsano:
Correct me if I'm wrong (I'm still wrapping my head around the code), but currently staking/unstaking does not have any impact on the calculation, right?
Although I have no solid argument for it, I would go for FIFO staking, as this makes more sense to me. Similar to the discussions in https://github.com/provinzio/CoinTaxman/issues/86 regarding separate wallets and https://github.com/provinzio/CoinTaxman/issues/4 for deposits/withdrawals, I would assume that I cannot sell a coin that I staked. Therefore, in your example, I would also sell the second coin first.

Griffsano:
What do you think of moving staked coins to a virtual "wallet/exchange" in order to "block" them? For example, when a coin on Kraken is staked, it is transferred from exchange "kraken" to virtual exchange "kraken-staked", and when it is unstaked, it is transferred from "kraken-staked" back to "kraken".
[...]

PR #100, opened 2022-01-02
Kraken: Action Types in CSV Exports

Griffsano:
This PR addresses #97 (but does not resolve all points mentioned in the issue).

@oldgitdaddy
Feel free to use this version for your Kraken exports. Let me know if you run into any problems.
If you have trades in USD, you can add this line in `kraken_pair_map` in `core.py`:
```
    "USDEUR": "USDTEUR",
```
The virtual sell feature may not work properly right now, so I recommend to disable `CALCULATE_VIRTUAL_SELL` in the `config.ini`. The problem is that no valid trades may be returned from the API for the current timestamp. Alternatively, a workaround for this is to use an older timestamp in https://github.com/provinzio/CoinTaxman/blob/main/src/taxman.py#L227.

I introduced a new `refids` list to track duplicate deposits/withdrawals and the additional deposit/withdrawal entries for staking / unstaking / staking reward actions. By tracking the Ref-IDs, duplicate entries can also be correctly identified when e.g. multiple deposits arrive in a short period of time (I think "overlapping" deposits/withdrawals could have been a problem before).
With this implementation, all first entries of deposits/withdrawals are ignored, and only the second entry leads to an operation. This way, the additional deposit/withdrawal lines for staking / unstaking / staking rewards are ignored.

The question is which deposit/withdrawal to ignore for "real" deposit/withdrawal actions? The current approach in the main branch considers the events on the blockchain as taxable (first deposit / second withdrawal). However, this could lead to erroneous FIFO/LIFO calculations e.g. if trades occur during the deposits/withdrawals (see #57). As I understand, the assets are only available for trading after the second deposit and before the first withdrawal.
@shredEngineer What do you think?


oldgitdaddy:
Thanks @Griffsano! I will be happy to check the changes, whenever they are in master. Can anybody please merge the PR?
Thanks!

provinzio:
As you mentioned the error, I would appreciate if you could test this branch with your full data set @oldgitdaddy

oldgitdaddy:
Ok. I will do the testing today and let you know.

oldgitdaddy:
The fix appears to work. It generated the tax evaluation for 2021. From my side this PR can be merged. Thanks!

Griffsano:
Update: The mapping from USDEUR to USDTEUR is not required anymore, since the inverse coin pair EURUSD is fetched now. Additionally, virtual sells should work now.

src/book.py
```diff
@@ -522,35 +508,54 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
-                    skip = dup_state[_type] == dup_skip[_type]
-                    dup_state[_type] = (dup_state[_type] + 1) % 2
-                    if skip:
+                # Skip duplicate entries for deposits / withdrawals and
+                # additional deposit / withdrawals lines for
+                # staking / unstaking / staking reward actions
+                if _type in ["deposit", "withdrawal"]:
+                    if refid not in refids:
+                        refids.append(refid)
                         continue
```
provinzio, line 517: Does the behavior changed? Before this PR the second deposit and the first withdrawal entry needed to be skipped. Now its always the first which gets skipped. Is this correct?

Could it happen, that there are more than two transactions with the same refid, so that the first gets skipped and the other ones will run through?

src/book.py
```diff
@@ -522,35 +508,54 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
 
                 # Parse data.
                 utc_time = datetime.datetime.strptime(_utc_time, "%Y-%m-%d %H:%M:%S")
                 utc_time = utc_time.replace(tzinfo=datetime.timezone.utc)
                 change = misc.force_decimal(_amount)
+                # remove the appended .S for staked assets
+                _asset = _asset.replace(".S", "")
```
provinzio, line 524: Is this safe or should we change it to `_asset.removesuffix(".S")` (which is actually nothing other than `_asset[:-2] if _asset and _asset.endswith(".S") else _asset`)?

src/book.py
```diff
@@ -569,13 +574,11 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
-        )
-        assert dup_state["withdrawal"] == 0, (
-            "Orphaned withdrawal. (Must always come in pairs). "
-            "Is your file corrupted?"
-        )
+        if margin_warnings:
+            log.warning(
+                f"{file_path}: {margin_warnings} margin entries. Margin trading is "
+                "currently not supported. Please create an Issue or PR."
+            )
```
provinzio, line 581: Shouldn't this be an error as this fucks up the entire tax calculation? Or is margin trading completly separated from normal trading?

src/price_data.py
```diff
@@ -361,6 +363,18 @@ def _get_price_kraken(
 
                 if not data["error"]:
                     break
+                elif data["error"] == ['EGeneral:Invalid arguments']:
+                    # cancel if inverse pair has been tried
+                    if inverse:
+                        num_retries = 0
+                        continue
```
provinzio, line 370: This is hard to read. Either add an comment (explicitly state, that we want to use for for-else case) or raise the error directly. We can adjust the error message and say, that normal and inverse pair was tried.

src/price_data.py
```diff
@@ -361,6 +363,18 @@ def _get_price_kraken(
...
+                    # try inverse pair
+                    else:
+                        log.debug("Invalid arguments error, trying inverse coin pair")
+                        inverse = not inverse
+                        temp_asset = quote_asset
+                        quote_asset = base_asset
+                        base_asset = temp_asset
```
provinzio, line 377: You can use `base_asset, quote_asset = quote_asset, base_asset` instead

src/price_data.py
```diff
@@ -391,8 +405,12 @@ def _get_price_kraken(
             if closest_match_index == -1:
                 continue
 
-            # The desired timestamp is in the future
-            if closest_match_index == len(data_timestamps_ms) - 1:
+            now_timestamp = misc.to_ms_timestamp(datetime.datetime.now().astimezone())
+            # The desired timestamp is in the future. Ignore if the target timestamp
+            # is very close to the current timestamp (for virtual sells)
```
provinzio, line 410: `            # is +/- 1 hour to the current timestamp (for virtual sells).`

src/price_data.py
```diff
@@ -391,8 +405,12 @@ def _get_price_kraken(
...
+            if closest_match_index == len(data_timestamps_ms) - 1 and \
+                (now_timestamp > target_timestamp + 3600 * 1000
+                    or not config.CALCULATE_VIRTUAL_SELL):
```
provinzio, line 413: I find this really hard to read.

What do you think of something like this?
```
# Check if he have to adjust the fetched timestamp interval.
# We have to adjust, if the fetched timestamp is in the future.
timestamp_in_future = closest_match_index == len(data_timestamps_ms) - 1
# We do not have to adjust, for future prices, when we just want to calculate virtual sells.
accept_timestamp = config.CALCULATE_VIRTUAL_SELL and abs(now_timestamp - target_timestamp) < 60 * 60 * 1000
if timestamp_in_future and not accept_timestamp:
    ....
```

Why can we accept the fetched price for virtual sells? This price is written to the database and used next year for tax calculation. I don't understand why prices for virtual calculations can be fetched differently. Please explain your thoughts.

provinzio (commented):
Hey @Griffsano,
I wish you a Happy New Year. :)
You did a great job, providing a hotfix for our user!

I noted some points in the code and have a question on how you treat price-fetching for virtual sells.
I am looking forward to your response, so that we can merge this asap.

src/book.py
Griffsano: Yes, this behavior changed, and I'm not sure yet what the best approach is here (I mentioned it in the PR text).

There are always two entries for deposits/withdrawals: One for crediting the transfer to/from the account, and one for writing it to the blockchain (according to the comments in the code). The action visible to others is the first deposit and second withdrawal. However, for the user of the account, the second deposit and first withdrawal may be more relevant (also for taxing), since this is when the user actually has the assets at their disposal. I think this is especially relevant for the FIFO/LIFO principle (example: you trade an asset while a deposit/withdrawal is ongoing).

What would you suggest here? If we want to consider the first deposit/withdrawal, we could create a "hold" list of operations, and only release them once the second deposit/withdrawal occurs.

For the other actions (staking / unstaking / staking rewards) that lead to "fake" deposits/withdrawals in the CSV exports, the deposits/withdrawals can just be ignored in my opinion (refer to the tables in #97).

Regarding your question, I've never seen more than two transactions with the same refid, except for margin trading.

src/book.py
Griffsano: Good point, I think this depends if the margin positions are closed or settled, and if closed, what the base for the trade was. If the gain/loss was payed in fiat/EUR at the end of the margin trade, then it should be fine. I need to think about it.
This leads to the question: Are gains/losses in fiat currency actually taxable? Let's say I trade BTC/EUR, then BTC/USD, then BTC/EUR, is the change in USD/EUR between the trades also relevant for taxation?

src/price_data.py
Griffsano: My thought was that for virtual sells you just want to get the current price, irrelevant if it's before or after the timestamp. But you're right, I didn't consider that this price is then stored and could mix up future tax calculations (if a trade is done at the same time as the virtual sell timestamp).

src/book.py
Griffsano: To answer my own question: Yes, gains/losses in other fiat currencies are taxable (refer to e.g. https://www.iww.de/astw/archiv/--23-estg--fremdwaehrungsgeschaefte-fuehren-zur-steuerpflicht-f51260)

Griffsano: To be on the safe side, I think it's best to tackle margin trading together with #52 and throw an error for now.

src/book.py
provinzio: I share your opinion. We should use the user-side transactions for tax calculations.

Do the two transactions differ in timestamps? I thought they are exact the same, but with different types. If that's the case, we might want to assert that the coin pair, value and timestamp are the same, before discarding it regardlessly.

Griffsano: Yes, the timestamps for deposits/withdrawals differ (couple of minutes). I've implemented a "hold list" for deposit/withdrawal operations and only release them after assertion when the second action occurs. It works well for the logs that I have, but please review if this implementation also makes sense to you. It would be nice to support multiple export files with the same "hold list", but this would require more implementation effort.

src/book.py
```diff
@@ -452,30 +452,15 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
-        #       Withdrawal is requested in Kraken account
-        #       > Skipped
-        #   dup_state["withdrawal"] == 1:
-        #       Withdrawal is broadcast to blockchain
-        #       > Taxable event (is in public trade history)
-        dup_state = {"deposit": 0, "withdrawal": 0}
-        dup_skip = {"deposit": 1, "withdrawal": 0}
+        # Need to track state of duplicate deposit/withdrawal entries
+        # All deposits/withdrawals are held back until they occur a second time
+        held_operations = []
```
Griffsano, line 463: Not sure how the annotation for `held_operations` should be. Linter error:
```
src/book.py:463:9: error: Need type annotation for 'held_operations' (hint: "held_operations: List[<type>] = ...")  [var-annotated]
src/book.py:631:41: error: Incompatible types in assignment (expression has type "object", variable has type "Optional[str]")  [assignment]
src/book.py:632:40: error: Incompatible types in assignment (expression has type "object", variable has type "datetime")  [assignment]
src/book.py:633:40: error: Incompatible types in assignment (expression has type "object", variable has type "str")  [assignment]
src/book.py:634:38: error: Incompatible types in assignment (expression has type "object", variable has type "Decimal")  [assignment]
src/book.py:635:36: error: Incompatible types in assignment (expression has type "object", variable has type "str")  [assignment]
src/book.py:636:35: error: Incompatible types in assignment (expression has type "object", variable has type "int")  [assignment]
src/book.py:637:41: error: Incompatible types in assignment (expression has type "object", variable has type "Path")  [assignment]
src/book.py:638:35: error: Incompatible types in assignment (expression has type "object", variable has type "Decimal")  [assignment]
```

src/price_data.py
Griffsano: So the problem is that the received price data is only accepted if the target timestamp is between the timestamps of the received data points. For virtual sells, the target timestamp (basically the current time) is usually newer than the latest available trade, which leads to rejection of the whole dataset.
I've implemented a different approach now: The Kraken API returns the latest 1000 trades. If less than 100 trades are received, it is assumed that the latest trade price has been reached, which is then returned. There's also a check that issues a warning if the target timestamp is older than one hour in such a case.
What do you think of this approach?

src/book.py
provinzio: does it work with `held_operations: list[dict[Str, Any]]]`? We would replace any with union[x,y,z] but any should be enough. Any is part of the typing module

Griffsano: Thanks, it worked now with `held_operations: list[dict[str, Any]] = []`.
However, for some reason the `str` type needs to be forced later on: `operation = str(held_operations[idx]["operation"])`. Else (without `str(...)`), this mypy error is raised:
`src/book.py:644:25: error: Argument 1 to "append_operation" of "Book" has incompatible type "Optional[str]"; expected "str"  [arg-type]`

src/price_data.py
provinzio: not perfect, but should be good enough for our use case

src/book.py
```diff
@@ -560,22 +561,99 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
-            "Orphaned withdrawal. (Must always come in pairs). "
-            "Is your file corrupted?"
-        )
+                    if fee != 0:
+                        self.append_operation(
+                            "Fee", utc_time, platform, fee, coin, row, file_path
+                        )
+                    if operation == "StakingInterest":
+                        # the rewarded coins are added to the staked portfolio
+                        # TODO: directly stake the rewarded coins, not just any coin
```
provinzio, line 658: Why should the rewarded coins directly be staked? Does Kraken work like this? Please write this as comment. I find this none trivial.

I am unsure about this, but staking might follow FIFO. So the current behavior is correct and the TODO is unnecessary.

provinzio (changes requested):

src/book.py
provinzio: see conversation https://github.com/provinzio/CoinTaxman/pull/100#discussion_r800065730

src/book.py
Griffsano: Yes, the way it works with Kraken is that the rewarded coins are directly added to the staked balance. Assuming that spot and staking are considered two separate "wallets", this is different from adding the rewards to the spot account and then moving the same amount to staking, because FIFO (staking of an older spot coin) would mess up the tax calculation.
That's why I added the todo. Following our discussion in https://github.com/provinzio/CoinTaxman/issues/57, it would mean that we have to add the rewards directly to the (virtual) `kraken-staked` exchange.

provinzio: Ah ok, that does make sense.

Keep in mind, that this should only be done when `MULTI_DEPOT` is activate. We might want to write this to the TODO.

src/book.py
```diff
@@ -560,22 +561,99 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
+                            operation = str(held_operations[idx]["operation"])
+                            utc_time = held_operations[idx]["utc_time"]
+                            platform = held_operations[idx]["platform"]
+                            change = held_operations[idx]["change"]
+                            coin = held_operations[idx]["coin"]
+                            row = held_operations[idx]["row"]
+                            file_path = held_operations[idx]["file_path"]
+                            fee = held_operations[idx]["fee"]
+                        else:
+                            append_operation = False
```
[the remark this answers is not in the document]
Griffsano: No, the refid is appended to `held_operations` list above (line 609). `append_operation = True` then adds the operation (line 649).

First and second withdrawal are both added to `held_operations`, but when only when the second one occurs the data of the first one is appended to the operations list.

src/book.py
Griffsano: good point, I'll add the comments

src/book.py
```diff
@@ -560,22 +561,99 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
+                        "file_path": file_path,
+                        "fee": fee,
+                    })
+
+                    if operation == "Deposit":
+                        # append only the second deposit to the operations list
+                        if refid_idxs:
+                            append_operation = True
+                        else:
+                            append_operation = False
```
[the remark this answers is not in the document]
Griffsano: yep, that's cleaner

src/book.py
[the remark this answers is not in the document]
provinzio: Ah ok, got it now. Thank you for the clarification. I find this part difficult to understand.

src/book.py
```diff
@@ -560,22 +561,99 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
+                # assets at their disposal. The first deposit and second withdrawal are
+                # in the public trade history and are skipped.
+                # For staking / unstaking / staking reward actions, deposits /
+                # withdrawals only occur once and will be ignored.
+                if operation in ["Deposit", "Withdrawal"]:
+                    # search for refid in refids list
+                    refid_idxs = [
+                        idx for idx, op in enumerate(held_operations)
+                        if op["refid"] == refid
+                    ]
```
provinzio, line 583: `idx` is never used alone, but always combined with `held_operations[idx]`.

So this list could also be `matching_operations = [op ....]` or something like this.

```
if len(matching_operations) == 0:
    # Matching operation wasn't found
    matching_operation = None
elif len(matching_operations) == 1:
    matching_operation = matching_operations[0]
else:
    log.error(...)
    raise RuntimeError
```

btw, this might be faster, when we use a dictionary instead: `dict[ref_id, ...]`.

provinzio: For the dict approach we need to save something with at least three states (not seen id, seen id, second id). We could use bool for that (None, False, True).

When we see the first deposit, we set it to False. When we see the second deposit, we set it to True and add the operation.

When we see the first withdrawal, we add the operation and set False. When we see the second withdrawal we set True.

Afterwards, we check wether all values in the dictionary are true or raise a warning, that we are missing a specific refid.

Griffsano: Hey @provinzio, I'm not sure if can follow you completely, I have some follow-up questions:

> `idx` is never used alone, but always combined with `held_operations[idx]`.
> So this list could also be `matching_operations = [op ....]` or something like this.

The "real operation item" op is only created afterwards when `append_operation` is called (not to be mixed with the `operation` variable, which is `"Deposit"` or `"Withdrawal"`). Therefore, `matching_operations` would need to be a subset of the `held_operations` list, but I don't see any advantage in that? Instead, we could introduce a temporary variable `matching_operation = held_operations[idx]` after checking the length of `idx` to prevent the repetition of `held_operations[idx]`?

> btw, this might be faster, when we use a dictionary instead: `dict[ref_id, ...]`. We can then use `held_operations.get(refid)` and check against None, wether we already saved something.

When you say that we could use a dictionary instead, do you mean a dictionary of dictionaries?:
```
held_operations = {
  {<refid1>: {'operation': <operation_name1>, 'utc_time': <utc_time1>, ...},
  {<refid2>: {'operation': <operation_name2>, 'utc_time': <utc_time2>, ...},
  ...
}
```

> As we do not want to save more than one operation, I think we do not need to save a list of operations in the dictionary, but just the relevant information like a string "first deposit found" or just True and for the withdrawal the previous operation (or add the withdrawal directly, but save it in the dict, to check if we found the second one.. might be better.

The `held_operations` contains the held-back operations of the whole CSV. It could be that between two deposits/withdrawals with the same refid there is another deposit/withdrawal action, so we need to keep track of all "open" held-back operations.
Additionally, we need to store all relevant information to append the operation at a later point in time (when the second withdrawal occurs, release/append the first withdrawal operation). This also allows us to compare the data of the first and second deposit/withdrawal.

> For the dict approach we need to save something with at least three states (not seen id, seen id, second id). We could use bool for that (None, False, True).

Do you want to store this as a separate variable (see below, all should be True at the end), or as part of `held_operations`?
```
refid_check = {
  <refid1>: None,
  <refid2>: False,
  <refid3>: True,
  ...
}
```

provinzio: Hey @Griffsano,


My comment on `matching_operations` doesn't matter if we implement `refid_check`. What i meant is, instead of saving `held_operations: list[refid]` we could save `held_operations: list[op]`. Because we nether user the refid for anything, we actually want the operations.
If we want to improve the performance, we can make it a `dict[refid, list[op]` to increase the speed of finding the correct entry.

----



I believe, that we do not need to save `held_operations`. It might be enough to save `refid_check` as you already described it. If we find an suitable operation in the loop, we just add it directly instead of saving it into the `held_operations`.

If we want to save only the first. No problem, just add it; it doesn't harm for now to have it added; we can still raise an error later, if the second couldn't be found.

If we want to save the second; just wait for it.

The `refid_check: dict[refid, bool]` dictionary is the key to all that. This dictionary will tell, if we found the first or second item.
```
None: not found (or not existing in the dictionary)
False: one found
True: two found
if we find another entry: raise error, thats not supposed to happen
````

Afterwards check `all(v == True for v in refid_check.values())`. If not all ids could be found twice, raise an error with the conflicting error.

Comment (provinzio) [reply src/book.py:583]: Ah, I see. In that case, it might be the easiest way to keep the `held_operations` variable. 

Preprocessing the list (check which ref_ids are duplicated and select the correct operations accordingly, is difficult to implement for different files in our current setup.

Then, we might profit from using a dictionary instead of a list (`dict[ref_id, operations]`)

Comment (provinzio) [reply src/book.py:583]: We can add the `held_operations` variable as Book class member (in `__init__`) or as class variable (directly after `class Book:`).

**Comment by Griffsano** [comment]
Hey @provinzio, I've refactored the logic using a class variable `defaultdict[str, defaultdict[str, Any]]`. In the nested defaultdict, there's now a a flag for storing if the operation has been appended already (None, False, True). In addition, only the data of the first operation is stored now (it's not necessary to store all operations, since only two should occur anyways).
I'm not so familiar with defaultdict (or with Python at all, for that matter :D). Is this what you had in mind?

Additionally, I've made some changes to the `get_price_kraken` function: Now pairs that returned an invalid arguments error are stored and not tried again, effectively cutting the API calls for flipped pairs in half.

**Comment by provinzio** [comment]
Awesome idea to remember the invalid pairs. 

defaultdicts are dictionaries which automatically initialize values for new keys. 

Let me show you this weird example, why defaultdicts can make your code cleaner. This is a function which creates a dict with keys 0 to 2. All keys store values from 0 to 2 which gets added incrementally. 

Doing this with `dict` you have to check first, wether the key is already in the dictionary. Otherwise you have to create the list first. 
```
d: dict[int, list] = {}
for i in range (3): for j in range(3):
  if i not in d:
    d[i] = []
  d[i]append(j)
```
A defaultdic toes exactly that for you. You don't have to check whether list key already exists in d. If a new key is added to a defaultdicts. It starts by creating a new value depending on the factory function (in this case `list()`)
```
import collections
d: dict[int, list] = collections.defaultdict(list)
for i in range 3: for j in range(3):
  d[i].append(j)
```

File: src/book.py [hunk 13]
```diff
@@ -574,73 +577,79 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
                 # in the public trade history and are skipped.
                 # For staking / unstaking / staking reward actions, deposits /
                 # withdrawals only occur once and will be ignored.
+                # The "appended" flag stores if an operation for a given refid has
+                # already been appended to the operations list:
+                # == None: Initial value, this is the first occurence
+                # == False: No operation has been appended, this is the second occurene
```
provinzio, line 583: Occurrence

src/book.py
```diff
@@ -574,73 +577,79 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
-                    # if refid already exists, assert that data of operations agree
-                    if refid_idxs:
-                        idx = refid_idxs[0]
+                    # If this is the first occurence, set the "appended" flag to false
+                    # and don't append the operation to the list. Instead, store the
+                    # data for verifying or appending it later.
+                    if self.kraken_held_ops[refid]["appended"] is None:
+                        append_operation = False
+                        self.kraken_held_ops[refid]["appended"] = False
+                        self.kraken_held_ops[refid]["operation"] = operation
```
provinzio, line 592: If you save all operation data as one key e.g. `d[refId]["operation"] : tuple or dict` you can easily discard the data as soon as we find the second value too free some memory. (~#overoptimization :D)

src/book.py
```diff
@@ -574,73 +577,79 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
+                    if append_operation and operation == "Withdrawal":
+                        # required for type annotation: convert Optional[str] to str
+                        operation = str(self.kraken_held_ops[refid]["operation"])
+                        utc_time = self.kraken_held_ops[refid]["utc_time"]
+                        platform = self.kraken_held_ops[refid]["platform"]
+                        change = self.kraken_held_ops[refid]["change"]
+                        coin = self.kraken_held_ops[refid]["coin"]
+                        row = self.kraken_held_ops[refid]["row"]
+                        file_path = self.kraken_held_ops[refid]["file_path"]
+                        fee = self.kraken_held_ops[refid]["fee"]
```
provinzio, line 652: If we use tuples to store the operation data and tuple unpacking (`a, b =tuple([1,2])`), we can make sure, that we do not forget to update this line of code in the future, when we add more operation data. 

Btw why don't we store the operation as real operation class? I think that it would look a lot nicer instead of always passing each value in it's own. 

We have to create a new help function for this, which creates the only the operation class an does not add it to the list of operations directly. So we call `create_operation` and use `self.operstions.append(op)` when we want to add the stored operation. 

What do think about that?

Griffsano: Hmm, it sounds cleaner but I don't know if it really makes things easier. If we add more items to the operation in the future, we have to revisit this place anyways (be it for append_operation or create_operation). Additionally, we would need to store an operation for the fee as well, effectively needing twice the memory due to the overhead of storing timestamp, platform etc. again. Right now, the fee is just an additional number. Furthermore, we would need an additional logic to decide between deposit/withdrawal and any other operation when actually appending the operation.

provinzio: We already have the logic to determine the operation type. Its the same that append operation uses

provinzio: Yeah we need to reserve more memory, but I find the implementation way more clearer, that I find it worth

Griffsano:
I've made the changes with storing the operation classes. However, I'm not happy yet with the implementation of the new create_operation / append_created_operation functions.

provinzio:
Append operation function should use the new create operation function. 

I propose that we distinguish between create/append operation function which get the raw data and create/append the operation and an helper function _append_operation which is nothing else than self.operstions.append

src/book.py
```diff
@@ -612,44 +608,66 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
...
+                        if operation == "Withdrawal":
+                            op = self.kraken_held_ops[refid]["operation"]
+                            op_fee = self.kraken_held_ops[refid]["operation_fee"]
+                        # Finally, append the operations and delete the stored
+                        # operations to reduce memory consumption
+                        self.append_created_operation(op)
+                        if op_fee:
+                            self.append_created_operation(op_fee)
+                        self.kraken_held_ops[refid]["operation"] = None
+                        self.kraken_held_ops[refid]["operation_fee"] = None
```
provinzio, line 670: del function can delete the key entirely from the dict

Griffsano:
Like this? There are still some mypy errors, not sure how to resolve them. The problem is that we want to return any subclass of Operation:
```
src/book.py:74:9: error: Returning Any from function declared to return "Type[Operation]"  [no-any-return]
src/book.py:104:32: error: Argument 1 to "_append_operation" of "Book" has incompatible type "Type[Operation]"; expected "Operation"  [arg-type]
src/book.py:665:52: error: Argument 1 to "_append_operation" of "Book" has incompatible type "Type[Operation]"; expected "Operation"  [arg-type]
Found 3 errors in 1 file (checked 12 source files)
```

provinzio:
Collections is a default python module

src/book.py
```diff
@@ -56,23 +56,23 @@ def create_operation(
         coin: str,
         row: int,
         file_path: Path,
-    ) -> Optional[tr.Operation]:
+    ) -> Type[tr.Operation]:
```
provinzio, line 59: This function does not return a type of operation but an object operation. The typing.type is too much.

src/book.py
```diff
@@ -56,23 +56,23 @@ def create_operation(
...
 
         try:
             Op = getattr(tr, operation)
```
provinzio, line 62: Mypy does not know what type getattr returns. Add an assert isinstance or in this case. Assert issubclass (dunno how the function is named. You might have to Google for this)

src/book.py
```diff
@@ -635,17 +632,16 @@ def _read_kraken_ledgers(self, file_path: Path) -> None:
                         try:
                             assert (
                                 isinstance(
-                                    type(op),
-                                    type(self.kraken_held_ops[refid]["operation"])
+                                    op, type(self.kraken_held_ops[refid]["operation"])
```
provinzio, line 635: You could also compare `op.__class__.name` with the string.

src/book.py
Griffsano: Then I'm getting
`src/book.py:74:9: error: Returning Any from function declared to return "Operation"  [no-any-return]`
even when asserting after `getattr`:
`assert issubclass(Op, tr.Operation)`

src/book.py
Griffsano: Do you think it would be faster?

src/book.py
Griffsano: Nevermind, works now, `assert isinstance(op, tr.Operation)` has to be done directly before return

src/book.py
provinzio: isinstance is typically used to check if object a is of type A. 

I thought about this again and we should use `type(op) is type(...)`. Mind the difference between is and ==. Might made now difference in our current setup, but it's more correct

Griffsano: Same that I got with ==
`./src/book.py:637:42: E721 do not compare types, use 'isinstance()'`

provinzio: Didn't expected that.... Than we might have to stock with the isinstance call. Although I find it quite weird used in this way. Whatever ¯\_(ツ)_/¯

Griffsano:
Alright, then we have closed all the open review items. I think it can be merged, what do you think?

provinzio:
Good job so far  👍 I am currently not at home and can not review your code properly. I'll set the review on my to-do list.

provinzio (approved):

PR #101, opened 2022-01-02
Add Bitpanda Exchange

### LINKED ISSUE #70: Bitpanda exchange

Griffsano:
Closes #70 
Based on #76


src/price_data.py
```diff
@@ -267,33 +269,41 @@ def _get_price_bitpanda_pro(
...
-                "period": t,
-                "from": begin.isoformat(),
-                "to": end.isoformat(),
-            }
-            r = requests.get(baseurl, params=params)
-
-            assert r.status_code == 200
+            # if no trades can be found, move 30 min window to the past
+            for o in range(0, 300, 30):
+                end = utc_time.astimezone(pytz.UTC) - datetime.timedelta(minutes=o)
```
provinzio, line 274: I think, that we do not need pytz here. Please have a look at [this stackoverflow answer](https://stackoverflow.com/a/2175170/8979290).

src/book.py
```diff
@@ -741,11 +741,168 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                    else:
+                        raise RuntimeError(
+                            f"Unknown asset class {asset_class}: Should be 'Fiat' or "
+                            f"'Cryptocurrency' in row {row} of file {file_path}"
+                        )
+                elif operation in ["Buy", "Sell"]:
+                    if asset_price_currency != config.FIAT:
+                        raise RuntimeError(
+                            "Only Euro is supported as 'Asset market price currency' "
+                            "currency, since price fetching for fiat currencies is not "
```
provinzio, line 850: remove `currency`

src/book.py
```diff
@@ -741,11 +741,168 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                            )
+                    elif asset_class == "Cryptocurrency":
+                        change = misc.force_decimal(amount_asset)
+                    else:
+                        raise RuntimeError(
+                            f"Unknown asset class {asset_class}: Should be 'Fiat' or "
+                            f"'Cryptocurrency' in row {row} of file {file_path}"
+                        )
+                elif operation in ["Buy", "Sell"]:
+                    if asset_price_currency != config.FIAT:
```
provinzio, line 847: You mixed up `config.FIAT` with Euro (in the error message, in `set_price_db`, and when adding Buy/sell operations).
I think it should be fine, if you replace all `Euro`/`EUR` with `upper(config.FIAT)`.

src/book.py
```diff
@@ -824,13 +981,48 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
+                    "Spread",
+                    "Spread Currency",
                 ],
             }
-            for exchange, expected in expected_headers.items():
-                if header == expected:
-                    return exchange
+            with open(file_path, encoding="utf8") as f:
+                reader = csv.reader(f)
+                # check all potential headers at their exprected header row,
```
provinzio, line 1017: `# Check all potential headers at their exprected header row.`

src/book.py
```diff
@@ -824,13 +981,48 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
+                # rewind the file after each expected header checked
+                for exchange, expected in expected_headers.items():
+                    header_row_num = expected_header_row[exchange]
+                    for _ in range(header_row_num + 1):
+                        header = next(reader)
+                        if header == expected:
+                            return exchange
+                    f.seek(0)
```
provinzio, line 1025: move comment here

`# Rewind the file after each header-check.`

src/book.py
```diff
@@ -824,13 +981,48 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
-                if header == expected:
-                    return exchange
+            with open(file_path, encoding="utf8") as f:
+                reader = csv.reader(f)
+                # check all potential headers at their exprected header row,
+                # rewind the file after each expected header checked
+                for exchange, expected in expected_headers.items():
+                    header_row_num = expected_header_row[exchange]
+                    for _ in range(header_row_num + 1):
+                        header = next(reader)
```
provinzio, line 1022: This raises an error for empty files or files which have less lines than `header_row_num`. Use `next(reader, None)` instead.

src/book.py
```diff
@@ -741,11 +741,168 @@ def _read_bitpanda_pro_trades(self, file_path: Path) -> None:
...
+                "binance": 0,
+                "binance_v2": 0,
+                "coinbase": 0,
+                "coinbase_pro": 0,
+                "kraken_ledgers_old": 0,
+                "kraken_ledgers": 0,
+                "kraken_trades": 0,
+                "bitpanda_pro_trades": 3,
+                "bitpanda": 6,
+            }
```
provinzio, line 905: File lines start with 1 not with 0. I would expect bitpanda row 6 to be in line 6 not in 7. I propose to increase the numbers by 1 for easier readability. `(header_row_num + 1)` below has to be adjusted then.

src/price_data.py
```diff
@@ -267,33 +269,41 @@ def _get_price_bitpanda_pro(
...
-                "unit": "MINUTES",
-                "period": t,
-                "from": begin.isoformat(),
-                "to": end.isoformat(),
-            }
-            r = requests.get(baseurl, params=params)
-
-            assert r.status_code == 200
+            # if no trades can be found, move 30 min window to the past
+            for o in range(0, 300, 30):
```
provinzio, line 273: Shouldn't this be (0, 30, 60) ?
We might want to adjust this for easier understanding to
```
num_retries = 0
for num_retries in range(3):
    # If no trades can be found, move window each time 30 min to the past
    o = num_retries * 30
    ....
```

provinzio (changes requested):
Hey @Griffsano,
thank you for finishing the work on #76 and progressing on #70.
I left my comments in the code.
Looking forward to your response.

Griffsano:
Hey @provinzio, happy new year! Thanks for the review, I updated the code accordingly.

src/price_data.py
```diff
@@ -267,33 +268,49 @@ def _get_price_bitpanda_pro(
...
+                params: dict[str, Union[int, str]] = {
+                    "unit": "MINUTES",
+                    "period": t,
+                    # convert ISO 8601 format to RFC3339 timestamp
+                    "from": begin.isoformat().replace('+00:00', 'Z'),
+                    "to": end.isoformat().replace('+00:00', 'Z'),
+                }
+                log.debug(
+                    f"Calling Bitpanda API for {base_asset} / {quote_asset} "
+                    f"price for {t} minute timeframe ending at {end}"
```
provinzio, line 294: Add information about `window_offset` in debugging message when `window_offset > 0`.

src/price_data.py
```diff
@@ -267,33 +268,49 @@ def _get_price_bitpanda_pro(
...
-                "from": begin.isoformat(),
-                "to": end.isoformat(),
-            }
-            r = requests.get(baseurl, params=params)
+                assert r.status_code == 200, "No valid response from Bitpanda API"
+                data = r.json()
 
-            assert r.status_code == 200
+                # exit loop if data is valid or maximum timeframe is not reached yet
+                if data or t != timeframes[-1]:
```
provinzio, line 302: Reading this from top to bottom, I thought at first, that we offset the window for each timeframe and got confused.

What do you think of this approach? It might be easier to understand when reading the code. We can update the comment above the for-loop to match our code.

```
# Try to find the price in the most detailed timeframe, to get the best matching price for the transaction.
# If we can not find the price in a timeframe, use the next bigger frame and try again.
# If we reached the highest timeframe, move the fetched time window into the past, to get the latest transaction from the API.
for t in timeframes:
  # Maximum number of allowed offsets into the past to find a valid price before we throw an error.
  # We do not offset the time window as long as we can choose a bigger timeframe instead.
  num_offsets = 6 if t == timeframes[-1] else 1
  for num_offset in range(num_offsets):
    ## attention: i changed the for-loop-variable name to singular
    ...
```

provinzio (changes requested):

src/price_data.py
Griffsano: yes, that's easier to read, I made the changes

src/price_data.py
```diff
@@ -267,33 +268,49 @@ def _get_price_bitpanda_pro(
...
+            for num_offsets in range(6):
+                # if no trades can be found, move 30 min window to the past
+                window_offset = num_offsets * 30
+                end = utc_time.astimezone(datetime.timezone.utc) \
+                    - datetime.timedelta(minutes=window_offset)
+                begin = end - datetime.timedelta(minutes=t)
+                if num_offsets:
+                    log.warning(
+                        f"No price data found for {base_asset} / {quote_asset} "
+                        f"at {end}, moving {timeframes[-1]} minute window to the past."
```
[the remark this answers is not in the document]
Griffsano: I moved the warning to the end of the loop (after the conditional break). Else, I get a linter error for uninitialized variable `end`.

provinzio:
Hey @Griffsano, fantastic work from you. I am gladly merging this into main.

Issue #104, opened 2022-01-23
Germany: Taxation of Airdrops


Griffsano:
According to https://www.winheller.com/bankrecht-finanzrecht/bitcointrading/bitcoinundsteuer/besteuerung-airdrops.html and the publication by the German BMF, Airdrops may be considered as "sonstige Einkünfte" in the future.

However, before updating the code accordingly, I think this topic needs some more investigation.

provinzio:
> However, before updating the code accordingly, I think this topic needs some more investigation.

What makes you  think that?

provinzio:
> However, before updating the code accordingly, I think this topic needs some more investigation.

What makes you  think that?

It might be difficult to differentiate between taxable airdrops ("irgendeine Leistung erbracht, bspw. personenbezogene Daten angegeben") and none taxable airdrops as descibe by Winheller. We might have to add an additional file, where the user can list all airdrops which shouldn't be count as taxrelevant.

Griffsano:
Exactly, the question is how to define if the user did something to get the airdrop. For example, some airdrops are distributed to investors achieving a certain trading volume within a given time frame. Would this also be a "Leistung"? The user probably would have traded anyways...

I like your approach with defining the tax-free airdrops in a file, but we have to define a format for this (similar to the deposits/withdrawals CSV-based matching idea https://github.com/provinzio/CoinTaxman/issues/4). Platform, UTC time and transaction ID if applicable should be sufficient to uniquely identify the airdrop. To be on the safe side, I would also consider all airdrops as tax relevant per default.

provinzio:
I revied the BMF file. `tax-fee` is kind of wrong. They differentiate between `Einkünfte aus sonstigen Leistungen` and `Schenkung`. I guess that no normal person gets over the `Freibetrag` for this, but we should identify it exactly like this in our tax report.

PR #115, opened 2022-03-26
Airdrop

Griffsano:
closes https://github.com/provinzio/CoinTaxman/issues/104

### DIFF
--- a/src/transaction.py
+++ b/src/transaction.py
@@ -108,7 +108,10 @@ class StakingInterest(Transaction):
 
 
 class Airdrop(Transaction):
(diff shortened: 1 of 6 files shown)


src/transaction.py
```diff
@@ -108,7 +108,10 @@ class StakingInterest(Transaction):
 
 
 class Airdrop(Transaction):
-    pass
+    taxable = True
```
provinzio, line 111: Default value needs to be defined in __init__. If you define it here. It's a general class variable. Alle classes share this variable

Issue #4, opened 2021-02-18
Resolve deposits and withdrawls


provinzio:
Foreign currency/coin deposits have to match withdrawls from another exchange. Otherwhise the program is not able to calculate the tax gains properly.

Furthermore we have to resolve deposits and withdrawals between the exchanges to correctly evaluate the taxation (#86).

We should check this in between the book read-in and the evaluation nd raise a warning or an exception. 

- [x] Find matching withdrawals and deposits or raise a warning
- [x] Resolve withdrawals/deposits from one exchange to another with FIFO/LIFO principle (according to config): Move transactions from the withdrawal-exchange to the deposit-exchange by changing `operations.platform`. Transactions might have to be splitted for this.

Adressed in https://github.com/provinzio/CoinTaxman/tree/4-resolve-deposits

Griffsano:
Just to confirm: Changing `operations.platform` would also maintain the price information for buying the coin, right?
Example: I buy a coin at exchange A for price 100. Then I transfer the coin to exchange B, where it has a price of 130 at the time of transfer. I sell the coin at exchange B for price 200. In this example, the taxable gain should be 100 and not 70 (minus of course the transfer/buy/sell fees).

provinzio:
Changing the platform might resolve in a different price. The price of a coin at buy time is dynamically evaluated with the `get_price` function. It checks the price of that coin at the buy time on the exchange `operations.platform`. So changing the platform changes the origin where the buy price comes from.

We might have to add an additional variable, so that we can distinguish between the buy-platform and the current platform.

Griffsano:
I agree, makes sense to add a new variable for the buy-platform and split up buy operations if necessary.

Already an idea for how to match deposits/withdrawals if the amount is not equal? (If I remember correctly, some exchanges directly withdraw the deposit fee before crediting the amount. -> We could back-calculate the fee by subtracting the deposit amount from the withdrawal amount.)
But there's always room for uncertainty, especially since transfers can take a while.

For deposits/withdrawals that cannot be matched, we could still move the respective buy operation to platform "unknown", so that at least the rest of the FIFO/LIFO tax calculation is correct.

provinzio:
If it's unclear, we have to rely on user input and issue a warning or a question to the user. We might want to create a `withdrawl-deposit-match.csv` in which you can write from where to where a coin was moved to explicitly define a withdrawl-match-pair.

Griffsano:
[...]

PR #116, opened 2022-03-26
Resolve Withdrawals/Deposits

Griffsano:
Closes https://github.com/provinzio/CoinTaxman/issues/4

### Dependencies that should be finished first

- [ ] Close PR https://github.com/provinzio/CoinTaxman/pull/98: This PR contains many changes in the balancing/taxation loop. Merging would be easier if this is done first. Once we separate balancing and taxation in two loops, we might get many merge conflicts and have to resolve them manually . However, not strictly required.

### To-do list

- [ ] Split up balancing and taxation in two different loops
- [ ] Add additional variable `acquire_platform` or similar to operations
- [ ] Change price fetching logic accordingly (use the correct `acquire_platform` or `platform`)
- [ ] Implement read-in of config files to match withdrawals and deposits (similar to https://github.com/provinzio/CoinTaxman/pull/115)
- [ ] Resolve deposits and withdrawals based on matching file
- [ ] Move coins between platforms, balancing for each platform (happens before taxation)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__
 account_statements/*
 data/*
 export/*
+withdrawals_deposits/*
 
 # Development folder
 dev/
--- a/src/book.py
+++ b/src/book.py
@@ -1107,6 +1107,31 @@ class Book:
                     overwrite=True,
                 )
 
+    def resolve_deposits(self) -> bool:
+        """ Matches withdrawals and deposits
+
+        Returns:
+            bool: Return True if everything went as expected.
+        """
+        paths = self.get_file_paths(config.WD_MATCHING_PATH)
+
+        if not paths:
+            log.warning(
+                "No matching files for withdrawals/deposits located in %s.",
+                config.WD_MATCHING_PATH,
+            )
+            return False
+
+        for file_path in paths:
+            #TODO match withdrawals/deposits
+            pass
+
+        if not bool(self):
+            log.warning("Unable to import any data.")
+            return False
+
+        return True
+
     def read_file(self, file_path: Path) -> None:
         """Import transactions form an account statement.
 
--- a/src/book.py
+++ b/src/book.py
@@ -1137,20 +1137,20 @@ class Book:
                 "Skipping file."
             )
 
-    def get_account_statement_paths(self, statements_dir: Path) -> list[Path]:
-        """Return file paths of all account statements in `statements_dir`.
+    def get_file_paths(self, folder_dir: Path) -> list[Path]:
+        """Return file paths of all input files in `folder_dir`.
+        For example, this function can return all account statement file paths.
 
         Args:
-            statements_dir (str): Folder in which account statements
-                                  will be searched.
+            folder_dir (str): Folder in input files will be searched.
 
         Returns:
-            list[Path]: List of account statement file paths.
+            list[Path]: List of input file paths.
         """
         file_paths: list[Path] = []
 
-        if statements_dir.is_dir():
-            for file_path in statements_dir.iterdir():
+        if folder_dir.is_dir():
+            for file_path in folder_dir.iterdir():
                 # Ignore .gitkeep and temporary excel files.
                 filename = file_path.stem
                 if filename == ".gitkeep" or filename.startswith("~$"):
@@ -1165,7 +1165,7 @@ class Book:
         Returns:
             bool: Return True if everything went as expected.
         """
-        paths = self.get_account_statement_paths(config.ACCOUNT_STATMENTS_PATH)
+        paths = self.get_file_paths(config.ACCOUNT_STATMENTS_PATH)
 
         if not paths:
             log.warning(
--- a/src/config.py
+++ b/src/config.py
@@ -59,4 +59,5 @@ BASE_PATH = Path(__file__).parent.parent.absolute()
 ACCOUNT_STATMENTS_PATH = Path(BASE_PATH, "account_statements")
 DATA_PATH = Path(BASE_PATH, "data")
 EXPORT_PATH = Path(BASE_PATH, "export")
+WD_MATCHING_PATH = Path(BASE_PATH, "withdrawals_deposits")
 FIAT = FIAT_CLASS.name  # Convert to string.
--- a/src/main.py
+++ b/src/main.py
@@ -40,8 +40,8 @@ def main() -> None:
     # 1) Find matching withdrawal and deposit or raise a warning
     # 2) Withdraw deposits from one exchange with FIFO/LIFO principle
     #    to the other exchange
-    # log.debug("Resolve withdrawals and deposits between exchanges...")
-    # book.resolve_deposits()
+    log.debug("Resolve withdrawals and deposits between exchanges...")
+    book.resolve_deposits()
     book.get_price_from_csv()
     taxman.evaluate_taxation()
     taxman.export_evaluation_as_csv()
--- a/src/main.py
+++ b/src/main.py
@@ -35,6 +35,13 @@ def main() -> None:
         log.warning("Stopping CoinTaxman.")
         return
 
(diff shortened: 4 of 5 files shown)


jhoogstraat:
If you provide a rough outline of how the solution is going to look, I am more than happy to help to implement this feature.

provinzio:
Hey  @jhoogstraat, awesome that you want to help. Does [this](https://github.com/provinzio/CoinTaxman/issues/4#issuecomment-1046324108)  give you the idea?

Griffsano:
Hey @jhoogstraat, thanks for helping! I added a to-do list in the top-comment above, based on the discussion in https://github.com/provinzio/CoinTaxman/issues/4. Let me know if this makes sense to you.
I think the most challenging part is resolving the deposits/withdrawals, as you need to consider all platforms simultaneously. Idea: Balance each platform until a deposit occurs, then pause and start/continue with the platform of the matching withdrawal (you can't continue after the deposit until you know which exact coin was transferred and where it originates).
I can implement the read-in of the deposits/withdrawals matching files, as I already have a concept for this and did something similar for the Airdrops.

jhoogstraat:
You can find the work in progress on my [fork](https://github.com/jhoogstraat/CoinTaxman/tree/resolve-deposits)
It does not work in the current state, but at least the general idea should be clear.
I must admit that with the limited time I had with the code, it is more difficult to understand than I thought.

Hopefully I got the distinction between "balancing" and "taxation" right. As I said, this is just a rudimentary implementation that is missing many things.

provinzio:
I think it should be possible to balance the platforms and evaluate the taxation at the same time. 

Is it necessary to add the new DepositMatch class instead of adding a new variable to all coins, which states where the coin is currently (besides the old variable for where the coin was bought)?

provinzio:
> I think it should be possible to balance the platforms and evaluate the taxation at the same time.
> 
> Is it necessary to add the new DepositMatch class instead of adding a new variable to all coins, which states where the coin is currently (besides the old variable for where the coin was bought)?

provinzio:
> Wasn't the goal to separate the balancing from taxation? Or what do you mean by "at the same time"?
> 
> Should the withdrawals/deposits better be evaluated beforehand (in `Book`, altering the list of operations?).
> 
> The new class is not strictly necessary. It currently just serves as a container for the rows in the wd.csv file. It allows to uniquely identify the related operations (deposit and withdrawal), so the operations can be matched.

Good points. Keep on the good work.

jhoogstraat:
Thanks!
Maybe I can clarify the thought process a bit more. Please correct me, if I am wrong.

The goal is for every operation to specify which coin exactly are exchanged via FIFO or LIFO queue. That would be equal to a `tuple[Operation, list[CoinSold]]` for each operation. 
That is the "balancing" in my understanding. After that the taxation goes though all tuples and can calculate the p/l of the operation.

That is basically how it works now, not respecting staking and potentially other operations.

provinzio:
I think balancing means the part when the buys/sells are evaluated. 

E.g. buy 1 coin. Buy 1 coin.  Sell 1 coin. Withdraw 1 coin. The withdrawal will move the second coin. Not the first. 

The idea of creating tuples per Operation is correct. It might be easier to copy the operation so that some coins stay with the old operation and some coins are part of the second operations which gets withdrawn.

E.g. buy 2 coins. Sell 1 coin. Sell 1 coin. Has operations buy(2), sell(1), withdraw(1). 
And another platform might have deposit(1), sell(1). 

The missing information is the link between the withdraw and deposit operation. So our first goal is to match deposit and withdraw operations and the second goal is to identify the coins with buy time and buy price. The information about the moved coins could be saved in the deposit operation. 

On tax evaluation. The withdraw operation will remove coins from the queue. The deposit operation will add/"buy" the coins which where identified earlier with the buy times and buy prices. This might be your tuple of sold coins. A tuple of operation and sold coins shouldn't be necessary as the sold coins know their operation (not sure about this)?

jhoogstraat:
Thanks for clarifying. 
I think my current implementation is not far off from the solution discussed in #4. 
I am evaluating the operation chain in a separate `evaluate_coin` method, *before* localized taxation kicks in.
By doing this, I was able to slim down the localized taxation method quite a bit, as it does not have to evaluate the chain itself, but just the profits/losses of each operation (its single responsibility) -> Fewer sources of errors.
As history cannot be changed, we could even cache the balancing to disk.

Withdrawals are handled as selling the coin on the specific platform queue, while deposits are just buys on the queue.
To evaluate withdrawals/deposits, it's just going through all operations and update the specific coins transferred for deposits.
Maybe that can be handled inside the first iteration through all operations, but I am not sure on that.

provinzio:
Saving the withdrawal/deposit links might be a good idea. But keep in mind, that I could add a missing account statement after the first run. So the saved link should also save the latest transaction. If an even older account statement file is added later on, the links might have to be recalculated.

provinzio:
some more ideas from me to the current implementation and my current favorite approach

I think that it's quite rare, that you have withdrawal/deposits in your history, which can not be matched automatically/easily

e.g.
deposit 5 btc requires a withdrawal of 5 btc + maybe fees

in my case: i rarly withdraw/deposit coins. for me it might be enough to just search for a deposit/withdrawal match by coin in an appropiate time frame.

using a file to match deposit/withdrawals is in my eyes over the top. i'd recommend to stick to a basic matching algo.

```
# get all deposit/withdrawal operations
# group by coin
# sort by time
```
If deposit/withdrawals come in the pattern withdrawal, deposit, withdrawal, deposit the matching should be trivial. (check the `change` to be sure)

jhoogstraat:
I agree. I'll implement the matching algo in a new branch first. After that we can implement the balancing part. Especially because you are refactoring the BalanceQueue right now.

provinzio:
> I agree. I'll implement the matching algo in a new branch first. After that we can implement the balancing part. Especially because you are refactoring the BalanceQueue right now.
Balancing won't be an issue as soon as the matching is done.

> Is it really enough to known the source platform when handling a deposit during balancing?
I'll recommend to save the Withdrawal operation to each Deposit. We'll require much more information than just the source platform. Also the handling will be much easier.

provinzio:
...and raise a warning when a Withdrawal matches to no Deposit. Make sure that each Withdrawal matches only to one Deposit, raise a warning when a Deposit has no Withdrawal.

jhoogstraat:
Does anyone have statements with some transfers in them? Doesn't matter if they are made up, as long as all transactions are valid.

It is definitely an issue that coinbase_pro and coinbase are treated as separate platforms, because withdrawals/deposits between them are not reported in the respective statements. Can we count them as one exchange from a legal standpoint? What exactly does qualify as a “separate platform”?

provinzio:
Difficult question. Might be enough to create your own transaction history. Make sure to account for withdrawal fees. E.g. withdrawal 100. Deposit 97

So we will need a matching Algo which looks for same coin and the same change or a bit lower change

jhoogstraat:
Maybe coinbase and coinbase_pro can be considered the same "platform" from a legal perspective? What makes a platform a platform?

Anyhow, yes, the change might be lower for the deposit, and we have to account for that. I don't know common fee rates, so maybe go with withdrawal change - 5% as a minimum?

provinzio:
Yeah why not. Just keep something like that or lower. 

If it's not enough. Issues will be raised. 
Perhaps start with 1%

jhoogstraat:
Done. How would you like to merge this?

provinzio:
Might be easiest when merged into my working branch instead of master.

jhoogstraat:
The match-fees branch? I can rebase and squash my changes onto the branch, so you can just cherry pick the commit. I can also create a new pr if you prefer.

provinzio:
Yes. 

Whatever you like most. You can reuse this branch if you like rebasing or resetting this.

provinzio:
In that case, just create a new branch.

PR #117, opened 2022-04-08
CHANGE price calculation for Coinbase and Bitpanda

Griffsano:
When Coinbase and Bitpanda CSV exports are read-in, the price is automatically stored in the database. The price is directly read-in from the CSV and may be rounded, which leads to warning messages when the price does not match the price stored in the database.

This PR changes this behavior, the price is now calculated instead of purely reading in the rounded price from the CSVs. This leads to more accurate price data and therefore less warning messages.

This is a follow-up PR to the discussion in https://github.com/provinzio/CoinTaxman/pull/106.

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -314,9 +314,9 @@ class Book:
                 utc_time = utc_time.replace(tzinfo=datetime.timezone.utc)
                 operation = operation_mapping.get(operation, operation)
                 change = misc.force_decimal(_change)
-                #  Current price from exchange.
-                eur_spot = misc.force_decimal(_eur_spot)
-                #  Cost without fees.
+                # Rounded price from CSV
+                # eur_spot = misc.force_decimal(_eur_spot)
+                # Cost without fees
                 eur_subtotal = misc.xdecimal(_eur_subtotal)
                 eur_fee = misc.xdecimal(_eur_fee)
 
@@ -326,8 +326,11 @@ class Book:
                 assert change
                 assert _currency_spot == "EUR"
 
-                # Save price in our local database for later.
-                set_price_db(platform, coin, "EUR", utc_time, eur_spot)
+                if isinstance(eur_subtotal, decimal.Decimal):
+                    # Calculated price
+                    price_calc = eur_subtotal / change
+                    # Save price in our local database for later.
+                    set_price_db(platform, coin, "EUR", utc_time, price_calc)
 
                 if operation == "Convert":
                     # Parse change + coin from remark, which is
@@ -933,7 +936,7 @@ class Book:
                 fiat,
                 amount_asset,
                 asset,
-                asset_price,
+                _asset_price,
                 asset_price_currency,
                 asset_class,
                 _product_id,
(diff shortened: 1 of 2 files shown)


provinzio:
Great idea to use other ways to calculate a more accurate price. Has the other column more decimal places? 

What happens when the `change` ist really high. Could it happen that at some point the old value is more accurate?

Griffsano:
> Great idea to use other ways to calculate a more accurate price. Has the other column more decimal places?

Yes, the other columns have the "real" unrounded paid/received values. Only the price columns are rounded.

> What happens when the `change` ist really high. Could it happen that at some point the old value is more accurate?

I don't have any really high/low price examples in my logs. But since the price column seems to be always rounded with two digits after the comma (regardless of how many decimals are before the comma), I think the calculated price should always be more accurate. In the end, this price is based on what you really paid/received (from the other unrounded columns), and is exactly what we use later for taxation.
For my tests with this method, I did not receive any database price matching warnings. This means that the prices calculated with the new formulas yield the same values as the `get_price_from_csv` function, which is based on the buy/sell operations.

> I'd prefer a way where we evaluate both "price types" and take the one with more decimal places or raising a warning if some decimal places are different (which shouldn't ever happen!)

Do you mean something like this?:
```
# price_column_price is raw price value read from CSV column
# calculated_price is calculated based on paid/received values
if abs(round(calculated_price, 2) - price_column_price) > 0.01:
    log.warning("...")
```
I'd allow a rounding error of 0.01 since we don't know if the prices in the CSV column is rounded up or down or to the closest 0.01 value.

provinzio:
> Yes, the other columns have the "real" unrounded paid/received values. Only the price columns are rounded.

Thank you for clarification.

> > I'd prefer a way where we evaluate both "price types" and take the one with more decimal places or raising a warning if some decimal places are different (which shouldn't ever happen!)
> 
> Do you mean something like this?:
> 
> ```
> # price_column_price is raw price value read from CSV column
> # calculated_price is calculated based on paid/received values
> if abs(round(calculated_price, 2) - price_column_price) > 0.01:
>     log.warning("...")
> ```
> 
> I'd allow a rounding error of 0.01 since we don't know if the prices in the CSV column is rounded up or down or to the closest 0.01 value.

Is not necessary in that case.

src/book.py
```diff
@@ -314,9 +314,9 @@ def _read_coinbase(self, file_path: Path) -> None:
                 utc_time = utc_time.replace(tzinfo=datetime.timezone.utc)
                 operation = operation_mapping.get(operation, operation)
                 change = misc.force_decimal(_change)
-                #  Current price from exchange.
-                eur_spot = misc.force_decimal(_eur_spot)
-                #  Cost without fees.
+                # Rounded price from CSV
+                # eur_spot = misc.force_decimal(_eur_spot)
+                # Cost without fees
                 eur_subtotal = misc.xdecimal(_eur_subtotal)
```
provinzio, line 320: Why is `xdecimal` (may be None) and `eur_spot` is always a decimal.

When `eur_splot` is always given, `eur_subtotal` should also be always given?

Griffsano: Yes, it is always given for my CSV logs. I also wondered about the same thing, but did not want to change `xdecimal` to `force_decimal` (it may have been written like that for a reason). We could change it, but I'm not sure if this breaks it for other CSV logs.

provinzio: They can open a bug request in that case ;) Change it to `force_decimal` for now.

provinzio (approved):

PR #64, opened 2021-08-16
Add Margin support

scientes:
This needs to be tested further

closes #52 

and i'm not sure if the gain/loss calculation is 100% correct because i'm not sure if we'd need to make seperate balance counters for margin and spot, but seperating those would be a major hassle

PR #120, opened 2022-04-10
Kraken Margin Trading

Griffsano:
Initial working draft for Kraken margin trading. Similar to @scientes' work in https://github.com/provinzio/CoinTaxman/pull/64.

Addresses https://github.com/provinzio/CoinTaxman/issues/52, https://github.com/provinzio/CoinTaxman/issues/97

### DIFF
--- a/src/taxman.py
+++ b/src/taxman.py
@@ -128,6 +128,7 @@ class Taxman:
                             transaction.CoinLendInterest,
                             transaction.StakingInterest,
                             transaction.Commission,
+                            transaction.MarginGain,
                         ),
                     )
                     and not sc.op.coin == config.FIAT
@@ -176,6 +177,33 @@ class Taxman:
             elif isinstance(op, transaction.Sell):
                 if tx_ := evaluate_sell(op):
                     self.tax_events.append(tx_)
+            elif isinstance(op, transaction.MarginFee):
+                balance.remove_fee(op.change)
+                if self.in_tax_year(op):
+                    # Fees reduce taxed gain.
+                    taxation_type = (
+                        "Kapitaleinkünfte aus Margin Trading (Werbungskosten)"
+                    )
+                    taxed_gain = -self.price_data.get_cost(op)
+                    tx = transaction.TaxEvent(taxation_type, taxed_gain, op)
+                    self.tax_events.append(tx)
+            elif isinstance(op, transaction.MarginGain):
+                balance.put(op)
+                if self.in_tax_year(op):
+                    taxation_type = "Kapitaleinkünfte aus Margin Trading"
+                    taxed_gain = self.price_data.get_cost(op)
+                    tx = transaction.TaxEvent(taxation_type, taxed_gain, op)
+                    self.tax_events.append(tx)
+            elif isinstance(op, transaction.MarginLoss):
+                # First, sell the lost coin to evaluate gain/loss from holding it
+                if tx_ := evaluate_sell(op):
+                    self.tax_events.append(tx_)
+                # Then, add the total loss to the income from capital
+                if self.in_tax_year(op):
+                    taxation_type = "Kapitaleinkünfte aus Margin Trading"
+                    taxed_gain = -self.price_data.get_cost(op)
+                    tx = transaction.TaxEvent(taxation_type, taxed_gain, op)
+                    self.tax_events.append(tx)
             elif isinstance(
                 op, (transaction.CoinLendInterest, transaction.StakingInterest)
             ):
--- a/src/transaction.py
+++ b/src/transaction.py
@@ -99,6 +99,18 @@ class Sell(Transaction):
     pass
 
 
+class MarginFee(Transaction):
+    pass
+
+
+class MarginGain(Transaction):
+    pass
+
+
+class MarginLoss(Transaction):
+    pass
+
+
 class CoinLendInterest(Transaction):
     pass
 
--- a/README.md
+++ b/README.md
@@ -208,3 +208,76 @@ Es ist also keine typische Kunden-werben-Kunden-Prämie sondern eher eine Kommis
 Für das Erste handhabe ich es wie eine Kunden-werben-Kunden-Prämie in Form eines Sachwerts.
 Sprich, die BTC werden zum Zeitpunkt des Erhalts in ihren EUR-Gegenwert umgerechnet und den Einkünften aus sonstigen Leistungen hinzugefügt.
 Aufgrund eines fehlenden steuerlichen Anschaffungsvorgangs ist eine Veräußerung steuerfrei.
+
+### Future- / Margin-Trading
+
+#### Unterscheidung Veräußerungsgeschäft / Termingeschäft
+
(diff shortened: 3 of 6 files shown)


provinzio:
Please add your sources of information, at least in the PR. So I can double check :)

Griffsano:
Hey @provinzio, I updated the README according to our discussion in #64:
https://github.com/Griffsano/CoinTaxman/blob/margin_kraken/README.md#future---margin-trading
This should contain all relevant information regarding my conception for margin taxation (if it's not clear, I'll try to clarify it directly in the README).

I'm still not sure if the tax calculation is 100% correct (especially since spot and margin wallets are combined for Kraken). However, at least the results for my exports make sense (profit/loss similar to Kraken dashboard).

provinzio:
Hey @Griffsano, thank you for your clarification. As I am currently reworking the evaluation and export, this PR is blocked by the branch `rework-evaluation-export`, but feel free to further work on the integration of Kraken Margin Trading.

Griffsano:
Hey @provinzio, I merged the current main branch into this PR and made the necessary changes so that margin evaluation for Kraken works again. Could you please review?

There's one thing that's weird though: Whenever I have margin entries in my CSV exports, the "Unrealisierte Einkünfte" part in the .xlsx just contains zeros, I haven't figured out yet why.

Btw, thanks for the detailed report feature, looks awesome :D

src/transaction.py
```diff
@@ -744,6 +764,10 @@ class StakingInterestReportEntry(InterestReportEntry):
     event_type = "Staking Einkünfte"
 
 
+class MarginReportEntry(InterestReportEntry):
+    event_type = "Margin-Trading"
```
provinzio, line 768: Does the labels from `InterestReportEntry` make sense here? All your MarginGain/MarginLoss.change are positive. So that the report should state that you always receive ("Erhalten am") coins. I believe that you have to update the labels. Also "Wert in EUR" doesn't make sense as we do not evaluate the value of the "bought" coins.

src/taxman.py
```diff
@@ -475,6 +477,54 @@ def _evaluate_taxation_GERMANY(self, op: tr.Operation) -> None:
                 )
                 self.tax_report_entries.append(report_entry)
 
+        elif isinstance(op, tr.MarginFee):
+            # Fees for margin trading
+            self.remove_from_balance(op)
+            if in_tax_year(op):
+                taxation_type = "Kapitaleinkünfte aus Margin Trading"
```
provinzio, line 484: Does this category really exists? I thought that these are normal "Kapitaleinkünfte". Please adjust or clarify.

src/taxman.py
```diff
@@ -475,6 +477,54 @@ def _evaluate_taxation_GERMANY(self, op: tr.Operation) -> None:
...
+                report_entry = tr.MarginReportEntry(
+                    platform=op.platform,
+                    amount=op.change,
```
provinzio, line 487: Shouldn't this be negative as we are giving it aways? Could be good to assert the `fee` value in `book.py`.

src/taxman.py
```diff
@@ -475,6 +477,54 @@ def _evaluate_taxation_GERMANY(self, op: tr.Operation) -> None:
...
+                    coin=op.coin,
+                    utc_time=op.utc_time,
+                    interest_in_fiat=-self.price_data.get_cost(op),
+                    taxation_type=taxation_type,
+                    remark=op.remark,
```
provinzio, line 492: It might be smart to add a remark that this is a Margin Fee.

src/taxman.py
```diff
@@ -475,6 +477,54 @@ def _evaluate_taxation_GERMANY(self, op: tr.Operation) -> None:
...
+                self.tax_report_entries.append(report_entry)
+
+        elif isinstance(op, tr.MarginLoss):
+            # Losses from margin trading
+            self.remove_from_balance(op)
+            if in_tax_year(op):
+                taxation_type = "Kapitaleinkünfte aus Margin Trading"
+                report_entry = tr.MarginReportEntry(
+                    platform=op.platform,
+                    amount=op.change,
```
provinzio, line 519: Shouldn't this be negative as we are giving it aways?

provinzio (changes requested):
Hey @Griffsano,

good first start to implement the Margin Trading into CoinTaxman. I appreciate your effort in writing the README.md. It gives a clear overview on the topic.

One question: at which time will the fees of a MarginTrade be recognized as tax loss? Already at the time of the paying (like currently implemented) or after the margin trade is closed?

I am a bit confused by the statement from winheller, which says, that your "Werbungskosten" can not be used to reduce your tax.

>der Abzug der tatsächlichen Werbungskosten ist ausgeschlossen

After this they say, that costs will become tax relevant on sell (which indicates that we have to adjust the code and match margin fees to margin-closes.

> In folgenden Fällen sind die Kosten auch ab 2009 weiterhin abzugsfähig:
> - Veräußerungskosten und Kosten in Zusammenhang mit Termingeschäften werden bei der Veräußerungsgewinnermittlung nach § 20 Abs. 4 EStG berücksichtigt.

src/transaction.py
```diff
@@ -179,6 +179,26 @@ class Sell(Transaction):
...
+    This is already a taxable value, no buy/sell calculation required."""
+
+    pass
+
+
+class MarginLoss(Transaction):
+    """Losses from margin trading.
+    This is already a taxable value, no buy/sell calculation required."""
+
+    pass
```
provinzio, line 199: When possible, it might be nicer to have classes `MarginStart`, `MarginFee`, `MarginClose` instead.
`MarginClose` should have a `fees: list[MarginFee]` variable. The fees must be linked to the MarginClose for tax evaluation.

provinzio:
I did some small formatting and merged main into the branch, to keep it up to date. :)

src/taxman.py
Griffsano: Added remarks to all three operation types (fee, gain, loss).

src/taxman.py
Griffsano: Yes, is fixed now.
I added an assertion to all three operation types.

src/taxman.py
Griffsano: Yes, is fixed now.

src/transaction.py
Griffsano: Changed to `TaxReportEntry`.
"Wert in EUR" makes sense for gains/losses/fees received/paid in crypto coins. However, since it's the same as "Gewinn/Verlust in EUR", I removed the column.
Not sure about the date (as it could be gain or loss), what do you think about "Erhalten oder ausgegeben am"?

Griffsano:
Hi @provinzio,

I finally had some time to continue working on this issue.

> I am a bit confused by the statement from winheller, which says, that your "Werbungskosten" can not be used to reduce your tax.

Yes, I was confused by the same sentence. Maybe they meant this?
> Das bedeutet, dass auch Aufwendungen im Zusammenhang mit der Einnahmeerzielung nicht mehr gegen Nachweis als Werbungskosten berücksichtigt werden. Sie sind mit dem Sparerpauschbetrag von 801 Euro bei Alleinstehenden und 1.602 Euro bei Verheirateten abgegolten.
(Source: https://www.steuergo.de/ro/texte/2021/597/kapitalvermoegen)

> One question: at which time will the fees of a MarginTrade be recognized as tax loss? Already at the time of the paying (like currently implemented) or after the margin trade is closed?

Good point. The problem is, that in case of Kraken, we probably can't link opening and closing actions, or the fees associated with the margin trade, as they have different reference IDs (see https://github.com/provinzio/CoinTaxman/issues/97).

src/price_data.py
```diff
@@ -468,7 +468,9 @@ def _get_price_kraken(
 
                 if not data["error"]:
                     break
-                elif data["error"] == ["EGeneral:Invalid arguments"]:
+                elif data["error"] == ["EGeneral:Invalid arguments"] or data[
+                    "error"
+                ] == ["EQuery:Unknown asset pair"]:
```
Griffsano, line 473: Ugly formatting, but this is how `black` autoformats

provinzio: What happens, if you put brackets around like this and use black to autoformat `((a) or (b))`?

Griffsano: thanks, looks better now

PR #121, opened 2022-04-13
Resolve Withdrawals/Deposits

jhoogstraat:
This is the follow-up for #116.

Contains a simple algorithm which matches related withdrawals and deposits.

Deposits now have a `link` property containing a reference to the matching withdrawal.
The algorithm checks that:

1. the coin matches
2. the deposit change is between .99 and 1 times the withdrawal change (accounting for potential fees).

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -1236,7 +1236,6 @@ class Book:
         withdrawal_queue: list[tr.Withdrawal] = []
 
         for op in sorted_ops:
-            # log.debug(op.utc_time)
             if isinstance(op, tr.Withdrawal):
                 if not misc.is_fiat(op.coin):
                     withdrawal_queue.append(op)
--- a/src/book.py
+++ b/src/book.py
@@ -1251,8 +1251,8 @@ class Book:
                 except StopIteration:
                     log.warning(
                         "No matching withdrawal operation found for deposit of "
-                        f"{match.change} {match.coin} "
-                        f"({match.platform}, {match.utc_time}). "
+                        f"{op.change} {op.coin} "
+                        f"({op.platform}, {op.utc_time}). "
                         "The tax evaluation might be wrong. "
                         "Have you added all account statements? "
                         "For tax evaluation, it might be importend when "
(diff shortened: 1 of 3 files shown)


jhoogstraat:
I see rebasing did not work as expected. One sec.

src/book.py
```diff
@@ -1212,6 +1212,56 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
+                            f"({op.platform}, {op.utc_time})"
+                        )
+                    except StopIteration:
+                        log.warning(
+                            "No matching withdrawal operation found for deposit of "
+                            f"{match.change} {match.coin} "
+                            f"({match.platform}, {match.utc_time})"
+                        )
+
+        log.info("Finished matching")
```
jhoogstraat, line 1263: This could just as well be a debug log.

src/book.py
```diff
@@ -1212,6 +1212,56 @@ def detect_exchange(self, file_path: Path) -> Optional[str]:
...
+                <= deposit.change
+                <= withdrawal.change
+            )
+
+        withdrawal_queue: list[tr.Withdrawal] = []
+
+        for op in sorted_ops:
+            # log.debug(op.utc_time)
+            if isinstance(op, tr.Withdrawal):
+                if not misc.is_fiat(op.change):
```
provinzio: we also should check for other fiats. only `config.FIAT` hasn't to be monitored.

provinzio (changes requested):

src/book.py
jhoogstraat: What do you mean by other fiats? `misc.is_fiat` already checks if the currency is any possible currency in `Fiat`.

provinzio: Selling USD might be a taxable income for a German citizien (like selling BTC), because it's a foreign fiat currency. I think that, although this tool is for crypto taxation, we should notify about that and try to evaluate the taxes for that, too.

provinzio: what I meant by check other fiat currencies is, that we also should match deposit and withdrawals for other fiat currencies

provinzio:
Hey @jhoogstraat, thank you for your good and helpfull contribution. The deposit/withdrawal match is an important milestone for a correct tax evaluation.

PR #122, opened 2022-04-13
Manage dependencies using pip-compile

jhoogstraat:
~This cleans up `requirements.txt` and `requirements-dev.txt` to only include direct dependencies of the project.
I also used the opportunity to update the version constraints to allow updates up to the next major version. This can be reverted if requested. I don't see why we should not pull the latest patches automatically, though.~

Edit:
The direction of this pr changed and is now aimed at managing the dependencies better. The idea stems from [here](https://stackoverflow.com/a/65666949).


provinzio:
I see your point in maintaining a clean requirements list with only our direct dependencies. But it's considered a good practice for a reason, to keep all dependencies even the indirect dependencies in the file - so that everybody who is working with this has the same outcome.

Updateing your dependencies without checking might lead to aweful bugs and misbehaviors. Therefore, we should pin our dependencies and update them only when wanted.

Feel free to implement something like `pip-compile` which helps to maintain the requirements.txt
[[dependencies ]()](https://stackoverflow.com/a/65666949/8979290)

jhoogstraat:
I see the point in keeping all dependencies. I guess we first need tests to make sure updated packages work as expected.

Anyway, dependencies should resolve successfully, which they do not currently.
I get an error installing `typed-ast`, which is resolved by updating mypy to the latest version.

What do you think about updating just the dev dependencies, or just mypy?

provinzio (commented):
The subdir is a great idea!

This might not be relevant for us now, but we could make sure, that our builded base and dev requirements always compatible. See this https://github.com/jazzband/pip-tools/issues/1020

jhoogstraat:
What is the reasoning for closing this pr?

provinzio:
Oh sorry. This didn't happen on purpose. I renamed my branch and GitHub closed the PR.

jhoogstraat:
Sorry, didn't mean to delete the branch. Restored it now.

provinzio:
I can not change the target branch... are you able to change it. The option should be available, when you click on edit.

jhoogstraat:
Did you decide against this pr?

provinzio:
Argh, I hate this auto closing feature, when the branch gets deleted....

Sorry... thank you for your reminder.

> The subdir is a great idea!
> 
> This might not be relevant for us now, but we could make sure, that our builded base and dev requirements always compatible. See this [jazzband/pip-tools#1020](https://github.com/jazzband/pip-tools/issues/1020)

Do we want to add this feature, before we merge this?

jhoogstraat:
Yes, sure! 
Give me a sec

jhoogstraat:
Can you reopen?
I pushed a new commit to my branch.

provinzio:
I don't see an option to reopen... Please create a new PR on main

PR #127, opened 2022-05-04
Rework evaluation export

provinzio:
The tax evaluation and export is currently under major rework. Most of the tax evaluation is done, but could require some more explanatory comments. The export is basically done, but I am still improving the formatting etc.

I'd love to have a helping hand, reviewing the changes, pinpointing code which could be better and so on....
If you have the time, I'd really appreciate it!

@Griffsano @scientes @jhoogstraat 

--------------

- [x] check draft of BMF against the PR https://www.bundesfinanzministerium.de/Content/DE/Downloads/BMF_Schreiben/Steuerarten/Einkommensteuer/2021-06-17-est-kryptowaehrungen.pdf?__blob=publicationFile&v=2
- [x] check that utc times are correctly exported (from import to export)
- [x] change timezone for export according to config timezone; update labels of ReportEntries
- [x] add deposit/withdrawal of fiat to export excel
- [x] taxable_gain looks wrong for sell after 1 year
- [x] add export sheet with all buys in tax year (information only)
check against
- [x] https://www.youtube.com/watch?v=qV5nfFddCZU&t=1050s

----------------

Related #104


src/taxman.py
```diff
@@ -61,324 +78,606 @@ def __init__(self, book: Book, price_data: PriceData) -> None:
...
+            # For now we'll just add our bought coins to the balance.
+            self.add_to_balance(op)
+
+        elif isinstance(op, tr.Sell):
+            # Buys and sells always come in a pair. The selling/redeeming
+            # time is tax relevant.
+            # Remove the sold coins and paid fees from the balance.
+            # Evaluate the sell to determine the taxed gain and other relevant
+            # informations for the tax declaration.
+            sold_coins = self.remove_from_balance(op)
```
Griffsano, line 334: Should this ignore `config.FIAT`? I'm getting the error `Not enough EUR in queue to sell`

src/main.py
```diff
@@ -39,16 +39,28 @@ def main() -> None:
...
     book.get_price_from_csv()
+    book.match_fees_with_operations()
+
     taxman.evaluate_taxation()
-    evaluation_file_path = taxman.export_evaluation_as_csv()
+    evaluation_file_path = taxman.export_evaluation_as_excel()
     taxman.print_evaluation()
 
     # Save log
     log_file_path = evaluation_file_path.with_suffix(".log")
```
Griffsano, line 58: This does not work if there is already a .log file with that name, but no .xlsx file. Shouldn't be a critical issue, but I just noticed it (I have older .log/.csv exports from working on the main branch, but no .xlsx files with the same name).

src/taxman.py
```diff
@@ -61,324 +78,606 @@ def __init__(self, book: Book, price_data: PriceData) -> None:
...
+                unrealized_sell = tr.Sell(
+                    utc_time=TAX_DEADLINE,
+                    platform=sc.op.platform,
+                    change=sc.sold,
+                    coin=sc.op.coin,
+                    line=[-1],
+                    file_path=Path(),
+                    fees=None,
+                )
+                self._evaluate_sell(
```
Griffsano, line 516: This is done independent from `CALCULATE_UNREALIZED_GAINS`

provinzio: totally forgot about the variable :rofl: thanks

src/main.py
provinzio: Next file name finder is now checking against multiple extensions (xlsx and log) which solves this problem.

src/main.py
```diff
@@ -39,16 +39,33 @@ def main() -> None:
         log.warning("Stopping CoinTaxman.")
         return
 
+    # Merge identical operations together, which makes it easier to match
+    # buy/sell to get prices from csv, match fees and reduces database access
+    # (as long as there are only one buy/sell pair per time,
+    # might be problematic otherwhise).
```
jhoogstraat, line 45: Noticed a small spelling mistake here.
```suggestion
    # might be problematic otherwise).
```

src/book.py
```diff
@@ -112,21 +121,26 @@ def _read_binance(self, file_path: Path, version: int = 1) -> None:
...
             "Commission Fee Shared With You": "Commission",
             "Referrer rebates": "Commission",
             "Referral Kickback": "Commission",
-            "Launchpool Interest": "StakingInterest",
-            "Cash Voucher distribution": "Airdrop",
-            "Super BNB Mining": "StakingInterest",
+            # DeFi yield farming
             "Liquid Swap add": "CoinLend",
             "Liquid Swap remove": "CoinLendEnd",
+            "Launchpool Interest": "CoindLendInterest",
```
jhoogstraat, line 137: Spelling mistake
```suggestion
            "Launchpool Interest": "CoinLendInterest",
```

jhoogstraat:
I don't know why this is throwing just now, but it does:
https://github.com/provinzio/CoinTaxman/blob/f076d8e31276daebbee2d89edb0f39338aae0521/src/book.py#L334

`_eur_subtotal`, `_eur_total` and `_eur_fee` are `None` when the operation is a withdrawal.
Here is an example from my coinbase statement:
`2021-01-01T00:00:00Z,Send,EOS,1,EUR,3,"","","","Sent 1,0000 EOS to binancecleos"`

jhoogstraat:
I am also getting many `Unable to correctly determine buy cost of bought coins...` warnings with my binance statement. I have to look into that.

provinzio:
> I am also getting many `Unable to correctly determine buy cost of bought coins...` warnings with my binance statement. I have to look into that.

Buys and sells always come in a pair (a trade). To determine the buy cost of bought coins, it is necessary to know the "sell"-part of the trade: the sell value is the buying cost of these coins.  This is important because (1) the bought coins value might be differ from the sell value and (2) the tax calculation requires you to use the real buying costs for tax calculation (see BMF draft)

Example why is is important:
- I buy a coin for 1 €, but the coin is at the same time worth 100 €. I just got lucky
- Now I am selling my coin for 100 €, which it is worth.

You can see, that I gained 99 €. This calculation is only possible, when Buy and Sell pairs got "matched" and the real buying cost of the coin (1 €) is used for tax calculation (100 € (buy value of the sell-trade) - 1 € (sell value of the buy-trade) = 99 €). If Cointaxman can not match the Buy/sell pairs it only sees

- Sell 1 € (worth 1 €)
- Buy 1 Coin (worth 100 €)
- Sell 1 Coin (worth 100 €)
- Buy 100 € (worth 100 €)

Stupid calculation of the gain calculates 100 € - 100 € = 0 € gain. This is what Cointaxman is currently doing for you, hence the warning. It does not realized which Buys and Sells form a trade-pair.


Now, how can we fix this? The matching is done in `book.py:resolve_trades`. Currently all operations are grouped by time. If at one time only a buy and sell trade exist: they belong together, chaka! If more than two operations were found, CoinTaxman does nothing, bad.... leading to your warning in the end.

To solve this. We have to extend the functionality of `resolve_trades` so that it's able to match your trade-pairs together. This is where you come into play. Please check your account statements and see which trades (operations happend at the same time) could'nt been matched. Why weren't they matched and what could we do to match them?

provinzio:
> I don't know why this is throwing just now, but it does:
> 
> https://github.com/provinzio/CoinTaxman/blob/f076d8e31276daebbee2d89edb0f39338aae0521/src/book.py#L334
> 
> `_eur_subtotal`, `_eur_total` and `_eur_fee` are `None` when the operation is a withdrawal (and potentially deposit and others). Here is an example from my coinbase statement: `2021-01-01T00:00:00Z,Send,EOS,1,EUR,3,"","","","Sent 1,0000 EOS to binancecleos"`

fixed as part of #130, already merged into the branch

Issue #135, opened 2022-05-16
Resolving multiple trades


jhoogstraat:
`resolve_trades` currently only works when there is exactly one `Buy` and one `Sell` op for any particular `utc_time`.
This leaves many cases unhandled, which could result in miscalculated gains.
Here are three such cases:

1) Multiple buy/sell-pairs at a particular `utc_time`
2) Orders not being fulfilled at once, leading to `Buy`/`Sell` ops later on.
3) Commissions / Withdrawals / (Deposits?) increasing the op count (more than 2 ops in that case)

The third case can be resolved by filtering the irrelevant ops.

provinzio:
Looks like deposit of eur and direct conversion to BNB. Transaction related should be something like buy/sell

jhoogstraat:
Ok, you are right, it didn't work because of the Deposit, which was not excluded when resolving trades.

PR #136, opened 2022-05-16
Resolve ops in edge cases

jhoogstraat:
This is the pr for issue #135.

This is not finished, but I wanted to allow for feedback early on.

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -21,7 +21,7 @@ import decimal
 import re
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Counter, Optional
 
 import config
 import log_config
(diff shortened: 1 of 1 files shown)


jhoogstraat:
I merged the two methods for resolving fees and trades into one. 
This reduces duplication by quite a bit.
It also enables us to resolve fees easily when there are multiple buy/sell-pairs at any given `utc_time`.

I am not sure how to account fees paid in bnb when there is more than one buy op... I added a `BUG` to the code at the relevant line.

provinzio:
> I merged the two methods for resolving fees and trades into one.

Make sure that resolving fees and resolving trades can be something different. E.g. the fee could also belong to a deposit 

> I am not sure how to account for fees paid in bnb when there is more than one buy op... I added a BUG to the code at the relevant line. How can we detect which bnb-fee op belongs to which buy op?

Split the BNB fee according to the "value" of the sell. We can use the price in eur to estimate that

jhoogstraat:
> Make sure that resolving fees and resolving trades can be something different. E.g. the fee could also belong to a deposit

True. Maybe have fees done in multiple passes? What do you prefer / think is better?

> Split the BNB fee according to the "value" of the sell. We can use the price in eur to estimate that

Like calculating the expected fee and see which combination of  fee ops matches the calculated fee best? Fees can be quite different depending on volume on Binance. Don't know about other platforms.

provinzio:
> True. Maybe have fees done in multiple passes? What do you prefer / think is better?

Matching of deposit/withdrawals, coin lend/staking start and end, trading pairs and finally matching everything with fees should be possible in the same loop

> In that case we might have to group and match everything before the "merge of equal operations"

jhoogstraat:
Im am first going to concentrate on trades with fees. We can add lending/staking, etc. later.

provinzio:
Hey @jhoogstraat,

I went through the code and did some formatting. Didn't wanted to bother you with that. I merged the newest main into the branch, fixed the linting errors and added some more assert checks.

Changes look good to me. I'd love when you could proof read my commits. Is the PR ready to be merged? Looking forward to your reply.

Issue #138, opened 2022-05-17
Transfers between Coinbase and Coinbase Pro not considered


jhoogstraat:
The reports from Coinbase and Coinbase Pro do not explicitly state transfers between the two.
In case of `MULTI_DEPOT` being enabled, this leads to missing coin on either side.

I am unsure whether the two can be considered different exchanges at all.

provinzio:
When the information if the withdrawal/deposit is missing think the statement files we have to count them as the same wallet...

jhoogstraat:
If that is allowed by our tax laws. They are technically two separate wallets.

provinzio:
> If that is allowed by our tax laws. They are technically two separate wallets.

I cannot confirm that, but if we are unable to reconstruct the deposit/withdraw between coinbase and coinbase pro, we have no choice. Maybe check that again or write an email to the Coinbase support?

jhoogstraat:
Their support site says
> Zu diesem Zeitpunkt müssen internationale Kunden ihre internationalen Steuerpflichten anhand ihres Transaktionsverlaufs ermitteln. Wir können keine zusätzlichen Informationen bereitstellen und keine lokal erforderlichen Formulare in Ihrem Namen einreichen. Wir empfehlen Ihnen, die Hilfe eines Steuerberaters in Ihrer Nähe anzufordern.

I mailed then still.. lets see.

Another thing:
Coinbase lets you buy coins directly with FIAT without depositing them first. So in case of Coinbase we should count trades with FIAT as deposit+trade.

provinzio:
But it doesn't have do be deposit+trade?

jhoogstraat:
Today, you actually can. I am unsure whether you could always deposit fiat.

I am unsure how to distinguish buys with deposited money or direct buy.

provinzio:
**Regarding the transfers between coinbase and coinbase pro:**

> Although your Coinbase and Coinbase Pro accounts are linked, they have their own separate wallets. Transfers from Coinbase to Coinbase Pro are free. 

Coinbase and coinbase pro are separeted wallets. See https://help.coinbase.com/en/pro/managing-my-account/funding-your-account/how-to-transfer-funds-between-your-coinbase-pro-and-coinbase-accounts

> Just saw that the `Account` statement on Coinbase Pro does include deposits and withdrawals. I think we should use that instead of the `Fills` statement. Doesn't fix the issue on the Coinbase side though.

[...]

PR #139, opened 2022-05-18
Add deposit when buying with fiat

jhoogstraat:
This adds a deposit operation when buying coins with fiat on coinbase.
Also sets the platform from "coinbase_pro" to "coinbase" to remove issue regarding transfers between the two.

Tracking issue: #138

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -413,6 +413,18 @@ class Book:
                     # If it's a buy, add the corresponding sell to complement
                     # the trading pair.
                     elif operation == "Buy":
+                        # If coins were bought with fiat, first add a corresponding deposit
+                        if misc.is_fiat(_currency_spot):
+                            eur_total = misc.force_decimal(_eur_total)
+                            self.append_operation(
+                                "Deposit",
+                                utc_time,
+                                platform,
+                                eur_total,
+                                _currency_spot,
+                                row,
+                                file_path,
+                            )
                         assert isinstance(eur_subtotal, decimal.Decimal)
                         self.append_operation(
                             "Sell",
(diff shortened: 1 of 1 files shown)


src/book.py
```diff
@@ -435,7 +447,7 @@ def _read_coinbase_v2(self, file_path: Path) -> None:
         self._read_coinbase(file_path=file_path)
 
     def _read_coinbase_pro(self, file_path: Path) -> None:
-        platform = "coinbase_pro"
+        platform = "coinbase"
```
provinzio, line 450: As discussed in #138, instead of making them the same, we should consider the deposits/withdrawals between them.

src/book.py
```diff
@@ -413,6 +413,18 @@ def _read_coinbase(self, file_path: Path) -> None:
...
+                            eur_total = misc.force_decimal(_eur_total)
+                            self.append_operation(
+                                "Deposit",
+                                utc_time,
+                                platform,
+                                eur_total,
+                                _currency_spot,
+                                row,
+                                file_path,
+                            )
```
provinzio, line 427: As it's also possible to deposit fiat first and buy later on, we have to check against that and can not merely add a deposit action.

It might be ok to disable tracking of deposit and withdrawals with home fiat which might solve most of the problem.
We can add a config variable for that. and should raise a warning when we adjust the config variable while evaluating the coinbase account statement, which as a buy with fiat.

If it is still bothering we should add a file in parallel which lists all buys to coinbase and initialize there value as 0 (do not add deposit on default). The user can adjust the file so that deposits will be added on the second run.

The format of the could be ini (csv might be a good "normal" format, but editing csv with Excel fucks up the format most of the time which might create a headache for most of the users) and should be saved in the account statements folder.

provinzio (changes requested):
Hey @jhoogstraat, 

please have a look at my comments. I belive that we should dig deeper into the problem to fully solve it.

Issue #186, opened 2026-08-30
Add "Simple Earn Locked - Rewards Income" for Binance staking operation


sandsturm:
Can we please add another Binance staking operation? I got this error:
`2026-08-30 15:34:27,733 book         ERROR    Could not recognize operation='Simple Earn Locked - Rewards Income' from platform='binance' in file_path=PosixPath('/CoinTaxman/account_statements/binance.csv') row=9469. The operation type might have been removed or renamed. Please open an issue or PR.
`

sandsturm:
Sorry, there are more:

1. binance: Simple Earn Flexible - Rewards Income
2. binance: Launchpool Airdrop - User Claim Distribution
3. binance: Launchpool Subscription/Redemption
4. binance: DOT Slot Auction Staking
5. binance: DOT Slot Auction Redemption
6. binance: DOT Slot Auction Rewards
7. binance: Stablecoins Auto-Conversion
8. binance: Simple Earn Flexible - Internal Transfer
9. binance: Buy Crypto With Fiat
10. binance: Convert Fiat to Stablecoin Paysafe
11. binance: Crypto Box Refund
12. binance: Simple Earn Flexible - Internal Transfer
13. binance: Stablecoins Auto-Conversion
14. binance: Transfer Between Spot and Funding

provinzio:
Sure, would you mind to create a pull request?

sandsturm:
I can make the change in book.py, but I am not sure how some of these transaction types should be handled in the ledger.

provinzio:
Feel free to make a draft. ai might be of help to find the correct transaction type. I'll check it afterwards

sandsturm:
Done. I think most of the mappings are clear. I was not fully sure about "Crypto Box Refund". Maybe you can have a more detailed look on this one?

PR #187, opened 2026-09-01
Handle additional Binance transaction types

sandsturm:
Adds mappings for additional Binance transaction types

Related to #186

### DIFF
--- a/src/book.py
+++ b/src/book.py
@@ -143,7 +143,6 @@ class Book:
             "HODLer Airdrops Distribution": "Airdrop",
             "Token Swap - Distribution": "Airdrop",
             "Launchpool Airdrop - System Distribution": "Airdrop",
-            "Launchpool Airdrop - User Claim Distribution": "Airdrop",
             #
             "Savings Interest": "CoinLendInterest",
             "Savings purchase": "CoinLend",
@@ -161,6 +160,7 @@ class Book:
             #
             "BNB Vault Rewards": "CoinLendInterest",
             "Launchpool Earnings Withdrawal": "CoinLendInterest",
+            "Launchpool Airdrop - User Claim Distribution": "CoinLendInterest",
             #
             "Commission History": "Commission",
             "Commission Fee Shared With You": "Commission",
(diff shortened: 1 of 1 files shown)


provinzio:
Does this make sense to you?


| Binance operation | PR maps to | Should be | Why |
|---|---|---|---|
| `Launchpool Airdrop - User Claim Distribution` | `Airdrop` | `CoinLendInterest` | Reward for locking coins, not a gratuitous allocation. With the default `ALL_AIRDROPS_ARE_GIFTS = True` it is reported as a gift and drops out of taxable income, although it is § 22 No. 3 EStG. Also matches the existing `Launchpool Interest` mapping. |
| `Crypto Box Refund` | `Deposit` | ignore (skip the row) | Return of the user's own unclaimed coins, not an acquisition. As an unlinked `Deposit` it creates a fresh cost basis and restarts the one-year holding period, and it double-counts the balance if the outgoing row was already booked. |
| `Buy Crypto With Fiat` | `Buy` | sign-based: `Sell` if `change < 0` else `Buy` (add to the `Binance Convert` tuple) | Two-legged trade. Hard-coding `Buy` also books the negative fiat leg as a buy, so the EUR balance rises instead of falling and the crypto leg loses its `link` (cost-basis warning). |
| `Convert Fiat to Stablecoin Paysafe` | `Buy` | sign-based: `Sell` if `change < 0` else `Buy` (add to the `Binance Convert` tuple) | Same two-legged conversion problem as above. |

sandsturm:
Understood. I have adapted book.py according to your feedback. Please have a look.

provinzio:
Hi @sandsturm, thanks for the PR and the additional Binance transaction type mappings! The linter failure here isn't caused by your code — it's an outdated `requirements.txt` in this branch (pinned to `requests==2.33.0`, which requires Python ≥3.10). I just fixed that on `main` — if you update your branch with `main`, CI should turn green again. Thanks for your patience!

sandsturm:
Thank you very much. I will update my main branch.

PR #188, opened 2026-09-07
Fix linter CI: pin dependencies back to Python 3.9-compatible versions

provinzio:
## Summary
The linter CI (flake8/isort/mypy) was broken because `requirements.txt`/`requirements-dev.txt` had drifted to pin versions (`requests==2.33.0`, `black==25.12.0`, ...) that require Python >=3.10, while the project still targets **Python 3.9** (per README/Dockerfile) — `pip install` failed before any linter could run. This PR does not change the project's target Python version; it only fixes the dependency pins so they actually work on 3.9 again.

## Key Changes
- **Runtime dependencies** (`dependencies/requirements.in` / `requirements.txt`): `requests` 2.33.0 → 2.32.5 (2.33.0 requires Python >=3.10); `tzdata` bumped to the current `2026.3` release (pure data package, unrelated to Python version)
- **Dev dependencies** (`dependencies/requirements-dev.in` / `requirements-dev.txt`): rolled back to the newest versions that still support Python 3.9: black 25.12.0 → 25.11.0, flake8-bugbear 25.11.29 → 24.12.12, isort 7.0.0 → 6.1.0, pip-tools 7.5.2 → 7.6.1, mypy 1.19.0 → 1.19.1, type stubs bumped to current compatible releases
- **`setup.cfg`**: fixed a masked flake8 config bug — the `ignore = \` continuation with an inline comment fails flake8 7.3.0's stricter validation (ignore codes must match `^[A-Z]{1,3}[0-9]{0,3}$`); replaced with a plain `ignore = W503` and a regular comment
- **`.github/workflows/linter.yml`**: still runs on Python 3.9 (unchanged from before)

Both lockfiles were regenerated with `pip-compile` running under an actual Python 3.9 interpreter, matching the workflow documented in the README.

## Verification
In a fresh Python 3.9 environment: `pip install -r requirements.txt -r requirements-dev.txt` succeeds, and flake8/isort/mypy all run to completion (no more install or config-parsing failures).

Note: the current CI run on this PR still shows flake8/isort/mypy failures — those are **pre-existing code issues** on `main` (long lines, an unused `global`, an import-order nit, a mypy `union-attr` finding), unrelated to this PR's diff, which only touches CI/dependency/config files. They were previously masked because the linter couldn't even install its dependencies.

https://claude.ai/code/session_01QQk1srzfkNU39D55KaySmT


provinzio:
CI status on the current head (`da4db2c`): dependency install now succeeds, but flake8/isort/mypy each still fail. All of the reported findings are **pre-existing on `main`** and unrelated to this PR's diff (which only touches `.github/workflows/linter.yml`, `dependencies/*.in`, `requirements*.txt`, `setup.cfg`) — they were simply masked before because the linter couldn't install its dependencies at all. Since fixing them would touch files outside this PR's scope, I'm not pushing those changes here; flagging them with proposed patches instead:

**mypy** — `src/transaction.py:95`
```
error: Item "str" of "Union[type[Any], str, Any]" has no attribute "__args__"  [union-attr]
```
`actual_type` can be a plain `str` at that point (see the `elif isinstance(actual_type, str)` branch above), so `.__args__` isn't guaranteed to exist. A minimal fix: only access `__args__` on `typing._SpecialForm`, not on `field.type` after the `isinstance` narrowing:
```python
if isinstance(actual_type, typing._SpecialForm):
    actual_type = typing.get_args(field.type)
```

**flake8** — 7× `E501` (line too long)
- `src/main.py:74` — wrap the f-string
- `src/taxman.py:764-765` — wrap the string literal
- `src/transaction.py:274,277,285,288` — wrap the docstring lines in `first_local_time`/`second_local_time`

**flake8** — `src/misc.py:266` `F824 global __delayed is unused: name is never assigned in scope`
`wrapper()` mutates `__delayed[...]` (item assignment) but never rebinds the name itself, so `global` is a no-op here under newer flake8's stricter check. Fix: drop the `global __delayed` statement (line 266) — mutation doesn't need it.

**isort** — import order
- `src/book.py:28` — `from typing import Any, Optional, NamedTuple` → `from typing import Any, NamedTuple, Optional`
- `src/config.py:18-19` — swap `locale`/`datetime` import order (alphabetical: `datetime` before `locale`)

Happy to open a small follow-up PR for these if useful, or take a review pass here if you'd rather fold them into this one — your call.

---
_Generated by [Claude Code](https://claude.ai/code)_

provinzio:
Pushed the proposed patches (`3ef508a`) — mypy/flake8/isort are all green now.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #190, opened 2026-09-08
Binance CSV: Token conversions lead to errors


sandsturm:
My CSV has these two transactions. Binance converted GAL to G with a ratio of 1:60 on July 17, 2024.
```
83258:76366993,2024-07-17 15:13:59,Spot,Asset Recovery,GAL,-0.92483873,
83259:76366993,2024-07-17 16:14:47,Spot,Token Swap - Distribution,G,55.4903238,
```

This leads to an error: 
```
2026-09-08 19:14:44,302 balance_queue ERROR    Not enough GAL in queue to sell: missing 0.00158569 GAL (transaction from 2024-07-17 15:13:59+00:00 on binance, see binance_complete.csv lines [83258])
This can happen when you sold more GAL than you have according to your account statements. Have you added every account statement including these from last years and the all deposits of GAL?
	This error may also occur after deposits from unknown sources. CoinTaxman requires the full transaction history to evaluate taxation (when and where were these deposited coins bought?).
```

It could be possible to infer token conversions, but maybe it makes more sense to introduce a config file for this, e.g. 
```
TOKEN_MIGRATIONS = {
    "GAL": {
        "to": "G",
        "ratio": Decimal("60"),
        "effective_from": datetime(2024, 7, 15, tzinfo=timezone.utc),
    },
}
```

sandsturm:
Thanks for your feedback. I have updated the code and pushed it.

PR #191, opened 2026-09-11
Handle Binance token migrations with preserved tax lots

sandsturm:
This PR adds support for deterministic token migrations in Binance account statements, such as GAL→G, MC→BEAMX, KLAY→KAIA, FTM→S, EOS→A, and OMNI→NOM. Migration definitions are loaded from a dedicated `migrations.ini,` while `migrations.py` provides the parsing and configuration model. book.py now matches configured `Asset Recovery` and `Token Swap - Distribution` pairs by source/target asset, ratio, date, and delay window, while leaving unmatched cases on the existing fallback path. New migration operation types preserve the link between outgoing and incoming assets without treating the migration as a taxable sale. taxman.py carries the original FIFO lots, acquisition dates, and cost basis into the successor token, including partial later disposals. The implementation was tested with synthetic Binance data covering multiple migrations, unmatched recovery/swap entries, and a partial sale of migrated tokens.

Related to #190

### DIFF
--- a/migrations.ini
+++ b/migrations.ini
@@ -1,0 +1,46 @@
+; Known deterministic token migrations used to match Binance
+; "Asset Recovery" and "Token Swap - Distribution" entries.
+; Only migrations with a known target asset and fixed conversion ratio
+; should be added here.
+
+[MC_BEAMX]
+source_coin = MC
+target_coin = BEAMX
+ratio = 100
+effective_from = 2023-11-01T00:00:00+00:00
+max_delay_hours = 72
+
+[GAL_G]
+source_coin = GAL
+target_coin = G
+ratio = 60
+effective_from = 2024-07-15T00:00:00+00:00
+max_delay_hours = 72
+
+[KLAY_KAIA]
+source_coin = KLAY
+target_coin = KAIA
+ratio = 1
+effective_from = 2024-10-28T00:00:00+00:00
+max_delay_hours = 72
+
+[FTM_S]
+source_coin = FTM
+target_coin = S
+ratio = 1
+effective_from = 2025-01-13T00:00:00+00:00
+max_delay_hours = 72
+
+[EOS_A]
+source_coin = EOS
+target_coin = A
+ratio = 1
+effective_from = 2025-05-26T00:00:00+00:00
+max_delay_hours = 72
+
+[OMNI_NOM]
+source_coin = OMNI
+target_coin = NOM
+ratio = 75
+effective_from = 2025-09-29T00:00:00+00:00
+max_delay_hours = 72
--- a/src/book.py
+++ b/src/book.py
@@ -30,6 +30,7 @@ import transaction as tr
 from core import kraken_asset_map
 from database import set_price_db
 from price_data import PriceData
+from migrations import TOKEN_MIGRATIONS
 
 log = log_config.getLogger(__name__)
 
@@ -141,7 +142,7 @@ class Book:
             "Launchpool Airdrop": "Airdrop",
             "Megadrop Rewards": "Airdrop",
             "HODLer Airdrops Distribution": "Airdrop",
-            "Token Swap - Distribution": "Airdrop",
+            "Token Swap - Distribution": "TokenSwapDistribution",
             "Launchpool Airdrop - System Distribution": "Airdrop",
             #
             "Savings Interest": "CoinLendInterest",
@@ -194,7 +195,7 @@ class Book:
             "Transaction Revenue": "Buy",
             "Transaction Sold": "Sell",
             "Transaction Fee": "Fee",
-            "Asset Recovery": "Sell",
+            "Asset Recovery": "AssetRecovery",
         }
 
         with open(file_path, encoding="utf8") as f:
@@ -1688,6 +1689,107 @@ class Book:
                 op.remarks.append("Ziel der Auszahlung unbekannt")
 
         log.info("Finished withdrawal/deposit matching")
+
+    def resolve_token_migrations(self) -> None:
+        matched: set[int] = set()
+        replacements: list[tr.Operation] = []
+
+        for migration in TOKEN_MIGRATIONS:
+            sources = [
+                (idx, op)
+                for idx, op in enumerate(self.operations)
+                if isinstance(op, tr.AssetRecovery)
+                and op.platform == "binance"
+                and op.coin == migration.source_coin
+                and op.utc_time >= migration.effective_from
+            ]
+
+            for source_idx, source in sources:
+                expected_target = source.change * migration.ratio
+
+                candidates = [
+                    (idx, op)
+                    for idx, op in enumerate(self.operations)
+                    if idx not in matched
+                    and isinstance(op, tr.TokenSwapDistribution)
+                    and op.platform == source.platform
+                    and op.coin == migration.target_coin
+                    and source.utc_time <= op.utc_time
+                    <= source.utc_time + migration.max_delay
+                    and op.change == expected_target
+                ]
+
+                if len(candidates) != 1:
+                    continue
+
+                target_idx, target = candidates[0]
+
+                migration_out = tr.TokenMigrationOut(
+                    utc_time=source.utc_time,
+                    platform=source.platform,
+                    change=source.change,
+                    coin=source.coin,
+                    line=source.line,
+                    file_path=source.file_path,
+                    remarks=source.remarks,
+                )
+
+                migration_out.target_coin = target.coin
+                migration_out.ratio = migration.ratio
+
+                migration_in = tr.TokenMigrationIn(
+                    utc_time=target.utc_time,
+                    platform=target.platform,
+                    change=target.change,
+                    coin=target.coin,
+                    line=target.line,
+                    file_path=target.file_path,
+                    remarks=target.remarks,
+                )
+
+                migration_in.source_coin = source.coin
+                migration_in.ratio = migration.ratio
+
+                migration_out.link = migration_in
+                migration_in.link = migration_out
+
+                matched.update((source_idx, target_idx))
+                replacements.extend((migration_out, migration_in))
+
+        self.operations = [
+            op
+            for idx, op in enumerate(self.operations)
+            if idx not in matched
+        ]
+
+        self.operations.extend(replacements)
+
+        # Restore the existing CoinTaxman behavior for Asset Recovery /
+        # Token Swap operations that did not match a configured migration.
+        for idx, op in enumerate(self.operations):
+            if isinstance(op, tr.AssetRecovery):
+                self.operations[idx] = tr.Sell(
+                    utc_time=op.utc_time,
+                    platform=op.platform,
+                    change=op.change,
+                    coin=op.coin,
+                    line=op.line,
+                    file_path=op.file_path,
+                    fees=op.fees,
+                    remarks=op.remarks,
+                )
+
+            elif isinstance(op, tr.TokenSwapDistribution):
+                self.operations[idx] = tr.Airdrop(
+                    utc_time=op.utc_time,
+                    platform=op.platform,
+                    change=op.change,
+                    coin=op.coin,
+                    line=op.line,
+                    file_path=op.file_path,
+                    fees=op.fees,
+                    remarks=op.remarks,
+                )
 
     def get_price_from_csv(self) -> None:
         """Calculate coin prices from buy/sell operations in CSV files.
--- a/src/main.py
+++ b/src/main.py
@@ -44,6 +44,9 @@ def main() -> None:
     # (as long as there are only one buy/sell pair per time,
     # might be problematic otherwise).
     book.merge_identical_operations()
+    # Resolve configured token migrations before price extraction and trade matching,
+    # so migration legs are not interpreted as ordinary sells or airdrops.
+    book.resolve_token_migrations()
     # Resolve dependencies between withdrawals and deposits, which is
     # necessary to correctly fetch prices and to calculate p/l.
     book.resolve_deposits()
--- a/src/migrations.py
+++ b/src/migrations.py
@@ -1,0 +1,49 @@
+import configparser
+import dataclasses
+import datetime
+import decimal
+
+from pathlib import Path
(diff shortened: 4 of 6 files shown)


provinzio:
Thanks for this — the approach is sound and it fits the existing architecture well.
Introducing dedicated migration operation types and carrying the FIFO lots through
them is the right shape for this problem, and the `migrations.ini` split keeps the
coin-specific knowledge out of the code.

Only §1 and §2 below block the merge, and both are small. Everything after that is
substance I would like to see addressed, but it is not gating.

## 1. [blocker] The holding-period fix never takes effect

`src/taxman.py`, in `_evaluate_sell`:

```python
is_taxable = (
    origin_sc.op.utc_time + relativedelta(years=1)
    >= op.utc_time
)

# Taxable when sell is not more than one year after buy.
is_taxable = sc.op.utc_time + relativedelta(years=1) >= op.utc_time   # overwrites
```

The second assignment is the pre-existing line and it overwrites the new one, so
the holding period is still counted from the **migration** date — exactly what
this PR sets out to fix. Example: GAL bought in 2022, migrated to G in 2024, sold
in late 2024 is reported as taxable despite being held well over a year.

It also makes the report internally inconsistent, because `buy_utc_time` a few
lines below already uses `origin_sc.op.utc_time`. And because
`_evaluate_unrealized_sells` routes through `_evaluate_sell` as well, the stale
line affects the unrealized-gains column too, not just realized sells.

Deleting the second assignment is the whole fix.

## 2. [blocker] CI is red

- **flake8**: `src/taxman.py` (W293, blank lines containing whitespace, two
  occurrences), `src/migrations.py` (W292, no newline at end of file).
  `migrations.ini` is also missing its trailing newline.
- **isort**: `src/book.py` — `from migrations import ...` sorts before
  `from price_data import ...`; `src/migrations.py` — blank lines inside the
  import block.

`make format` fixes all of these except the `migrations.ini` newline — `black`
only runs over `src/`, so that one needs adding by hand.

One caveat: `src/book.py:278` (W291) is **already red on `main`** and is not your
doing, but flake8 runs over the whole repo, so this branch stays red until that
one is fixed too. Feel free to include it here.

mypy is green.

## 3. [should fix] Write down the tax assumption

This is not a code issue, and I will not hold the merge for it, but it is the
change I most want in this PR.

The PR makes a tax-law decision without saying so: it treats a token migration as
a non-taxable event and continues the original acquisition date. **I agree with
that decision** — a swap at a fixed, predetermined ratio is a conversion of the
same economic asset, not a disposal, so it does not restart the one-year holding
period under § 23 EStG. But nothing in the code or the docs tells a reader that
the decision was made, and the BMF-Schreiben on crypto assets has no explicit
rule for token migrations, which is precisely why the reasoning needs to be
written down rather than baked in silently.

- [ ] A README section stating that CoinTaxman treats these migrations as
      non-taxable with the acquisition date carried over, with the reasoning:
      economic identity (*Nämlichkeit*) is preserved, the closest analogy is a
      share split, there is no realisation act, and the BMF-Schreiben's
      *Fußstapfentheorie* for hard forks points the same way. Users can then
      raise it with their tax adviser if they disagree.
- [ ] A comment in `migrations.py` stating that only **deterministic, forced
      migrations with a fixed ratio** belong in `migrations.ini`. The header
      comment in `migrations.ini` hints at this but does not say why it matters.
- [ ] Ideally a config switch to turn the behaviour off without editing code. Not
      required, but see the report-entry point in §5.

Happy to write the README text myself if you would rather not — say the word.

## 4. [should fix] `max_delay_hours = 72` looks too short for most entries

I checked the six entries against the Binance announcements. Every **ratio is
correct**, and the `effective_from` dates line up with or precede the delisting
dates. The delay window is the problem:

| Migration | Ratio | Delisting (UTC) | New token tradable (UTC) | Gap |
|---|---|---|---|---|
| MC→BEAMX | 100 ✓ | 2023-11-07 03:00 | 2023-11-14 08:00 | **~173 h** |
| GAL→G | 60 ✓ | 2024-07-15 03:00 | 2024-07-19 08:00 | **~101 h** |
| KLAY→KAIA | 1 ✓ | 2024-10-28 03:30 | 2024-10-31 07:00 | **~75.5 h** |
| FTM→S | 1 ✓ | 2025-01-13 03:30 | 2025-01-16 07:00 | **~75.5 h** |
| EOS→A | 1 ✓ | 2025-05-26 03:00 | 2025-05-28 08:00 | ~53 h ✓ |
| OMNI→NOM | 75 ✓ | late Sept 2025 | — | unverified |

Important caveat: those are announcement timestamps for trading and deposit
opening, not necessarily the timestamps on the `Token Swap - Distribution` rows
in an account statement — the credit may well land earlier. You presumably have a
real statement, which I do not.

- [ ] Could you paste the actual `Asset Recovery` and `Token Swap - Distribution`
      timestamps you saw for one or two of these? If the real gap is above 72 h,
      the shipped defaults match nothing for those coins — and because of §5 they
      would do so silently.

Since the window is per-migration anyway, I would set it per entry from the
observed gap plus headroom rather than keeping one 72 h default.

One thing to keep while adjusting the window: the amount comparison should stay
exact. All configured ratios are integers, so `Decimal * Decimal` is exact and
there is no rounding to absorb — and a loosened amount match would let an
unrelated distribution pass as a migration, silently turning a genuinely taxable
airdrop into a non-taxable event. The other real failure mode is source rows
split across several timestamps, which `merge_identical_operations` will not
combine because it only merges identical timestamps. That would need aggregation
across the window rather than a looser comparison — I would rather see the
diagnostics below first and find out what real statements actually look like.

## 5. [should fix] Matching is completely silent

There is not one log line in `resolve_token_migrations`. When matching fails, the
user silently gets a taxable disposal instead of a migration, and the only way to
notice is to read the Excel output and spot that a coin vanished.

- [ ] Warn on every `AssetRecovery` that has a configured migration but no match,
      with enough detail to diagnose the pairing from the log alone — including
      the target distributions that were rejected *only* by the delay window, and
      how far off they were. Something like:

```python
log.warning(
    f"No token migration match for {source.change} {source.coin} "
    f"(Asset Recovery at {source.utc_time}, "
    f"{source.file_path.name} lines {source.line}).\n"
    f"\tExpected {expected_target} {migration.target_coin} "
    f"within {migration.max_delay} of it.\n"
    f"\tFound: "
    + ", ".join(
        f"{op.change} {op.coin} at {op.utc_time} "
        f"({(op.utc_time - source.utc_time).total_seconds() / 3600:.1f} h later)"
        for op in target_distributions  # all of them, window or not
    )
)
```

That way a user can tell "the distribution was there, just 101 hours later" from
"the amount was off" without touching a debugger — which is exactly what is
needed to settle §4.

- [ ] A closing `log.info("Finished token migration matching")`, consistent with
      `resolve_deposits`.

Related, and the same transparency concern: a matched migration produces **no
report entry at all**. In the Excel export, coins simply vanish and others
appear. Whatever view one takes on taxability, that is the part a tax office
would question. An informational entry (along the lines of `TransferReportEntry`)
would make the migration auditable.

## 6. [should fix] Cite the source of each ratio in `migrations.ini`

Each of these numbers is a tax-relevant constant, and the file currently gives no
way to check where any of them came from. Binance publishes a "Binance Has
Completed the … Token Swap" announcement per migration that states the ratio, and
those URLs are stable.

- [ ] Add the relevant announcement URL as a comment in each section, e.g.:

```ini
[GAL_G]
; 1 GAL = 60 G
; https://www.binance.com/en/support/announcement/...
source_coin = GAL
...
```

## 7. [should fix] Reject two migrations sharing a source coin

The candidate list filters on `idx not in matched`, but the `sources` list does
not. Two sections with the same `source_coin` — A→B and A→C — would therefore
consume the same operation twice and draw the balance down twice.

- [ ] Add `and idx not in matched` to the `sources` comprehension. One token, and
      it closes the hole regardless of what else lands.
- [ ] Reject the configuration when loading `migrations.ini`, with an error
      naming both sections — it is always a config error, not something the
      matching loop should be resolving.

(Chained migrations A→B→C are a different case and already work: the B leg
appears as two separate rows — an incoming `Token Swap - Distribution` and a
later outgoing `Asset Recovery` — so they pair up independently, and the `while`
loop in `get_origin_sold_coin` unwinds both levels. No change needed there.)

## 8. [should fix] `load_token_migrations()` has no validation and runs at import

`migrations.py` is imported by `book.py`, so a typo in `migrations.ini` surfaces
as a raw `KeyError` or `decimal.InvalidOperation` during import — a traceback
instead of a message, for a file that is explicitly meant to be edited by users.

- [ ] Wrap the load so any failure raises one readable error naming the file and
      the offending section.
- [ ] Require `source_coin`, `target_coin` and `ratio` to be present, and
      validate `source_coin != target_coin`.
- [ ] Validate `ratio`: parses as `Decimal`, greater than zero, and **integral**.
      The exact amount comparison relies on the ratio being integral (§4); if a
      fractional ratio is ever configured, exact equality stops being the right
      test, so it should not be accepted silently.
- [ ] Validate `max_delay_hours` is positive.
- [ ] Handle `effective_from` robustly. Two traps, both of which a user editing
      this file will hit: a value without an offset (`2024-07-15`) parses to a
      naive datetime and then raises
      `TypeError: can't compare offset-naive and offset-aware datetimes` against
      `op.utc_time`; and on Python 3.9, which this repo targets,
      `datetime.fromisoformat` rejects a trailing `Z` outright with `ValueError`.
      My suggestion is to accept both spellings and treat a naive value as UTC
      with a warning, rather than failing — it is only a lower bound for
      matching, so an hours-level misreading is harmless.

Optional: moving the `load_token_migrations()` call out of module scope and into
`resolve_token_migrations()` would make failures surface where they are
actionable rather than at import of `book.py`.

## 9. [nit] `migrations.ini` does not reach the Docker image

The Dockerfile copies only `requirements.txt`, `Makefile` and `src`, and
`configparser.read()` ignores a missing file without complaint, so
`TOKEN_MIGRATIONS` would end up empty in a container.

To be clear, this is **not your problem to fix**: `config.ini` is not copied
either, and `config.py` raises `NotImplementedError` at import when it is
missing, so the image as built cannot run at all today. I am noting it only so it
does not get lost — I will deal with the packaging separately. If you want to
copy `migrations.ini` alongside, fine; if not, also fine.

## 10. [nit] Smaller things

- `get_buy_cost` contains a second, recursive copy of the `get_origin_sold_coin`
  logic. `_evaluate_sell` is its only caller and already has `origin_sc` in hand,
  so passing `self.get_buy_cost(origin_sc)` and deleting the
  `isinstance(sc.op, tr.TokenMigrationLot)` branch from `get_buy_cost` is the
  simplest form. (Reusing the helper inside `get_buy_cost` works too, if you
  prefer the safety net.)
- The tolerance assert in the `TokenMigrationIn` branch is unreachable, and its
  comment claims the opposite of what the code does. `book.py` matched on exact
  equality, and `balance_queue.remove()` raises `RuntimeError` on an
  underfunded balance before the assert is reached, so `migrated_total` and
  `op.change` are always exactly equal here. Either drop it or make the comment
  match §4's reasoning.
- `TokenMigrationOut` / `TokenMigrationIn` / `TokenMigrationLot` are not listed
  in `operations_order`, so `sort_by_order_and_key` gives them `idx = 0` and
  sorts them before everything else at an equal timestamp. It happens to work
  today because the pair is appended in order and the sort is stable, but that is
  accidental — `TokenMigrationOut` belongs with the loss operations, the other
  two with the gains.
- `src/migrations.py` is missing the AGPL header that every other module in
  `src/` carries.
- `BASE_PATH` in `migrations.py` duplicates `config.BASE_PATH` — import it.
- `op.platform == "binance"` is hardcoded on the source side while the target
  side uses `op.platform == source.platform`. Since `AssetRecovery` and
  `TokenSwapDistribution` are only ever produced by the Binance reader, the check
  is redundant either way.
- `TokenMigrationOut` / `TokenMigrationIn` derive from `Operation`, not
  `Transaction`, and `match_fees` filters on `tr.Transaction`. A fee row sharing
  a timestamp with a migration would no longer be matched. Almost certainly
  irrelevant in practice, but worth being aware of.
- `migrations.ini` is not mentioned in the README at all.

---

To summarise the priorities: §1 and §2 are what block the merge and both are a
couple of lines. §3 (write the assumption down), §4 (does the window actually
match?) and §5 (make failures visible) are what I care about most in substance.
The rest is robustness and polish. Happy to take §3 off your hands — and thanks
again, this is a genuinely useful addition.

---
_Generated by [Claude Code](https://claude.ai/code)_

sandsturm:
I am doing the code changes and will share it soon. In the meantime, please find attached the actual timestamps from my Binance account statement. Based on these observed gaps, I changed max_delay_hours to 24h for MC/GAL/KLAY/OMNI and 48h for FTM/EOS, with some headroom rather than using a single 72h default.

```
MC:
2023-11-13 09:41:28  Asset Recovery               MC     -3.29132658
2023-11-13 12:07:07  Token Swap - Distribution    BEAMX  +329.132658
Gap: ~2h 26m

GAL:
2024-07-17 15:13:59  Asset Recovery               GAL    -0.92483873
2024-07-17 16:14:47  Token Swap - Distribution    G      +55.4903238
Gap: ~1h 01m

KLAY:
2024-10-30 08:32:54  Asset Recovery               KLAY   -0.26922624
2024-10-30 09:40:15  Token Swap - Distribution    KAIA   +0.26922624
Gap: ~1h 07m

FTM:
2025-01-13 19:12:48  Asset Recovery               FTM    -0.0126
2025-01-15 08:48:53  Token Swap - Distribution    S      +0.0126
Gap: ~37h 36m

EOS:
2025-05-27 07:14:46  Asset Recovery               EOS    -6.52067961
2025-05-28 08:22:44  Token Swap - Distribution    A      +6.52067961
Gap: ~25h 08m

OMNI:
2025-09-30 14:09:26  Asset Recovery               OMNI   -0.68989106
2025-10-01 04:21:17  Token Swap - Distribution    NOM    +51.7418295
Gap: ~14h 12m
```

sandsturm:
Just a small follow-up. I have incorporated the review feedback and the PR is ready for another review when you have time.

